PR #4, opened 2017-04-05
Return QR Code info on cloud build

Mitko-Kerezov:
Return QR code info on cloud build

Merge after https://github.com/NativeScript/nativescript-cli/pull/2691

Ping @TsvetanMilanov @rosen-vladimirov

### DIFF
--- a/lib/services/cloud-build-service.ts
+++ b/lib/services/cloud-build-service.ts
@@ -17,6 +17,7 @@ export class CloudBuildService implements ICloudBuildService {
 		private $projectFilesManager: IProjectFilesManager,
 		private $errors: IErrors,
 		private $server: CloudService.IServer,
+		private $qr: IQrCodeGenerator,
 		private $mobileHelper: Mobile.IMobileHelper,
 		private $projectHelper: IProjectHelper,
 		private $logger: ILogger) { }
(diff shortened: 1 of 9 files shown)


lib/services/service-proxy.ts
```diff
@@ -10,6 +10,7 @@ export class ServiceProxy implements CloudService.IServiceProxy {
 		path = `appbuilder/${path}`;
 		headers = headers || Object.create(null);
 		headers["X-Icenium-SolutionSpace"] = headers["X-Icenium-SolutionSpace"] || "Private_Build_Folder";
+		headers["User-Agent"] = `fusion/${require("../../package.json").version} (Node.js ${process.versions.node}; ${process.platform}; ${process.arch})`;
```
rosen-vladimirov: for version you can reuse the code from here https://github.com/NativeScript/nativescript-cloud/blob/master/lib/commands/cloud-lib-version.ts#L11-L15

lib/services/cloud-build-service.ts
```diff
@@ -59,14 +51,20 @@ export class CloudBuildService implements ICloudBuildService {
...
 
 		this.$logger.info(`The result of ${buildInformationString} successfully downloaded. Log from cloud build is:${EOL}* stderr: ${buildResult.Error}${EOL}* stdout: ${buildResult.Output}${EOL}* outputFilePath: ${localBuildResult}`);
 
+		const buildResultUrl = this.getBuildResultUrl(buildResult);
+
 		return {
 			stderr: buildResult.Error,
 			stdout: buildResult.Output,
-			outputFilePath: localBuildResult
+			outputFilePath: localBuildResult,
```
rosen-vladimirov: I believe the server returns the FullOutput as well, can we get it here as well?

PR #10, opened 2017-05-15
Use new server

TsvetanMilanov:
We have a new server which we are going to use for our cloud services. It has endpoints for authentication, cloud builds and cloud emulators. 
The public API of the cloud builds service is not changed. The logic in the cloud build service has some changes because the server returns different response. The new result is a JSON object with S3 links to the `build status`, `build output` and the `build result`. The new build logic will be:
- Send request to start the build. This will return the links to the S3.
- Check the build status to see if there is error or it is complete.
- Get the build output from the output link and emit / log it.
- When the build is finished download the build result.

This PR also adds a login and dev-login commands.
How login works:
- Start server on localhost with port `p`.
- Open browser with the login url and `p` in the query parameters.
- After a bunch of redirects we will end up on the login page.
- If the login is successful the server will return redirect to localhost:`p` with the login response data.
- Save the user data.
How dev-login works:
- Send directly the username and the password to the server and receive the login response data in the response body.
- Save the user data.

> Note: The dev-login command works only with Telerik accounts.

The server configuration is also changed. There are two types of server urls:
- Production urls - We will have subdomain for each service. We will create the address for each service based on the subdomain name and the domain name.
- PR urls - In each PR build we will produce full urls for each service. This urls won't have subdomain and base domain. They will be completely different urls for each service in each PR. That's why we will use them directly to call the server.

TODO: Discuss how to support build logs for multiple builds.
TODO: Add docs for the build output event.


lib/services/authentication-service.ts
```diff
@@ -0,0 +1,146 @@
...
+		let isResolved = false;
+
+		this.$logger.info("Launching login page in browser.");
+
+		let loginUrl: string;
+		let localhostServer = this.$httpServer.createServer({
+			routes: {
+				"/": async (request: ServerRequest, response: ServerResponse) => {
+					this.$logger.debug("Login complete: " + request.url);
+					let parsedUrl = parse(request.url, true);
```
rosen-vladimirov: const

lib/services/authentication-service.ts
```diff
@@ -0,0 +1,146 @@
...
+					let loginResponse = parsedUrl.query.response;
```
rosen-vladimirov: const

lib/services/authentication-service.ts
```diff
@@ -0,0 +1,146 @@
...
+			this.$logger.debug("Login URL is '%s'", loginUrl);
+
+			if (!options.openAction) {
+				this.$opener.open(loginUrl);
+			} else {
+				options.openAction(loginUrl);
+			}
+
+			if (!isInteractive()) {
+				let timeout = options.hasOwnProperty("timeout")
```
rosen-vladimirov: const

lib/services/authentication-service.ts
```diff
@@ -0,0 +1,146 @@
...
+			}
+		});
+
+		const loginResponse = await authComplete;
+
+		if (timeoutID !== undefined) {
+			clearTimeout(<NodeJS.Timer>timeoutID);
+		}
+
+		let userData: IUserData = JSON.parse(loginResponse);
```
rosen-vladimirov: const

lib/services/authentication-service.ts
```diff
@@ -0,0 +1,146 @@
...
+		this.$userService.setUserData(userData);
+
+		let userInfo = userData.userInfo;
```
rosen-vladimirov: const

lib/services/cloud-services/cloud-services-proxy.ts
```diff
@@ -0,0 +1,105 @@
...
+			requestOpts.body = theBody.value;
+			requestOpts.headers[HTTP_HEADERS.CONTENT_TYPE] = theBody.contentType;
+		}
+
+		let response: Server.IResponse;
+		try {
+			response = await this.$httpClient.httpRequest(requestOpts);
+		} catch (err) {
+			if (err.response && err.response.statusCode === 402) {
+				this.$errors.fail({ formatStr: "%s", suppressCommandHelp: true }, JSON.parse(err.body).Message);
```
rosen-vladimirov: failWithoutHelp please

README.md
```diff
@@ -103,6 +103,395 @@ tns.cloudBuildService
...
+	 * @returns {Promise<ITokenState>} Returns the token state
+	 */
+	getCurrentUserTokenState(): Promise<ITokenState>;
+}
+
+interface ILoginOptions {
+	/**
+	 * Action which will be used to open the login url.
+	 */
+	openAction?: (loginUrl: string) => void;
```
TsvetanMilanov: We don't need to await this promise because our logic does not depend on the opening of the login page. It depends on the redirect to the localhost server.

lib/services/cloud-build-service.ts
```diff
@@ -155,6 +171,74 @@ export class CloudBuildService implements ICloudBuildService {
 		}
 	}
 
+	private async getObjectFromS3File<T>(pathToFile: string): Promise<T> {
+		return JSON.parse(await this.getContentOfS3File(pathToFile));
+	}
+
+	private async getContentOfS3File(pathToFile: string): Promise<string> {
+		return (await this.$httpClient.httpRequest(pathToFile)).body;
```
TsvetanMilanov: We don't need this because the `body` is string.

PR #15, opened 2017-05-23
Change openAction return type

Mitko-Kerezov:
And bump version to `0.4.0`

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,12 +187,10 @@ const tns = require("nativescript");
 const childProcess = require("child_process");
 
 const openAction = url => {
-		return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


lib/services/authentication-service.ts
```diff
@@ -60,7 +60,7 @@ export class AuthenticationService implements IAuthenticationService {
 			if (!options.openAction) {
 				this.$opener.open(loginUrl);
 			} else {
-				options.openAction(loginUrl);
+				options.openAction(loginUrl).catch(reject);
```
rosen-vladimirov: We've discussed it with Tsvetie and we've agreed that openAction will not be a promise, so we do not need this catch. Just update the method definition and Public API please.

PR #18, opened 2017-05-26
Improve authentication service.

TsvetanMilanov:
We should add `cancelLogin` method because when we use the `nativescript-cloud` as a module in long living process we want to have a way to stop the current login session.

When we login our identity services sets a cookie. If we don't clear this cookie after logout we will automatically login the user on the next login request. We don't want this behavior so we should add request to `logout` endpoint on our server to clear the cookie. Because of this change we have a breaking change in our logout api - the logout method accepts optional options object with open action. This open action will be used to open the logout url.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,32 @@ tns.authenticationService.refreshCurrentUserToken()
 	.catch(err => console.error(err));
 ```
 
+* `cancelLogin` - Stops the current login process and rejects the login promise with an error.
+</br>
+
+Definition:
+
+```TypeScript
+/**
+ * Stops the current login process and rejects the login promise with an error.
+ * @returns {void}
+ */
+cancelLogin(): void;
+```
+</br>
+
+Usage:
+```JavaScript
+const tns = require("nativescript");
+
+tns.authenticationService
+	.login()
+	.then(userInfo => console.log(userInfo))
+	.catch(err => console.error(err));
+
+tns.authenticationService.cancelLogin();
+```
+
 #### Interfaces:
 ```TypeScript
 interface IUser {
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -293,9 +335,11 @@ interface IAuthenticationService {
 
 	/**
 	 * Invalidates the current user authentication data.
+	 * If options.openAction is provided, it will be used to open the logout url instead of the default opener.
+	 * @param {ILoginOptions} options Optional settings for the logout method.
```
Mitko-Kerezov: The JSDocs says the param is of `ILoginOptions` type and the actual method's arg is marked as `IOpenActionOptions`
I know one inherits the other but it'd still be nice to be consistent

lib/services/authentication-service.ts
```diff
@@ -23,19 +30,20 @@ export class AuthenticationService implements IAuthenticationService {
...
+		this.localhostServer = this.$httpServer.createServer({
 			routes: {
 				"/": async (request: ServerRequest, response: ServerResponse) => {
 					this.$logger.debug("Login complete: " + request.url);
 					const parsedUrl = parse(request.url, true);
 					const loginResponse = parsedUrl.query.response;
 					if (loginResponse) {
 						await this.serveLoginFile("end.html")(request, response);
-						localhostServer.close();
+						this.localhostServer.close();
```
Mitko-Kerezov: Do you think it appropriate to set `this.localhostServer = null;` here?
If so you can extract this functionality and use it in `cancelLogin` down below too

lib/services/authentication-service.ts
```diff
@@ -100,7 +109,33 @@ export class AuthenticationService implements IAuthenticationService {
...
+			this.rejectLoginPromiseAction(new Error("Login canceled."));
+			this.rejectLoginPromiseAction = null;
+		}
+	}
+
+	public logout(options?: IOpenActionOptions): void {
+		options = options || {};
+
+		const logoutUrl = this.$authCloudService.getLogoutUrl();
+		this.$logger.trace(`Logging out. Logotu url is: ${logoutUrl}`);
```
Mitko-Kerezov: Typo in text `Logotu`

PR #19, opened 2017-05-29
Implement clean build option

nadyaA:
* To trigger an incremental, build the build request data JSON
  must set the property useIncrementalBuild to "true". This is the default
  behavior. Using the --clean option will explicitly require
  clean & full build.
* Although the nativescript-cloud is an extension that is used only with
  nativescript projects, current implementation of the builder daemon will
  not add default framework. This breaks tooling when incremental build is
  performed. Passing the framework here is more consistent that adding conditional
  behavior in the tooling.

http://teampulse.telerik.com/view#item/319525

### DIFF
--- a/lib/services/cloud-build-service.ts
+++ b/lib/services/cloud-build-service.ts
@@ -236,7 +236,7 @@ export class CloudBuildService extends EventEmitter implements ICloudBuildServic
 		});
 	}
 
-	private async prepareBuildRequest(projectSettings: { projectDir: string, projectId: string, projectName: string, nativescriptData: any },
+	private async prepareBuildRequest(projectSettings: IProjectSettings,
 		platform: string, buildConfiguration: string): Promise<any> {
 
 		const projectZipFile = await this.zipProject(projectSettings.projectDir);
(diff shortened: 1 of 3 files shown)


nadyaA:
Can the client determine whether there's need for clean and full build?

nadyaA:
Would the --clean option be clear enough for the users? How they'll know, if for example they have new plugin, that they have to pass the option 💭

lib/services/cloud-build-service.ts
```diff
@@ -249,6 +249,10 @@ export class CloudBuildService extends EventEmitter implements ICloudBuildServic
 		const cliVersion = await this.getCliVersion(runtimeVersion);
 		const sanitizedProjectName = this.$projectHelper.sanitizeName(projectSettings.projectName);
 
+		/** Although the nativescript-cloud is an extension that is used only with nativescript projects,
+		 * current implementation of the builder daemon will not add default framework. This breaks tooling when incremental build is
+		 * performed. Passing the framework=tns here is more consistent that adding conditional
+		 * behavior in the tooling.*/
```
rosen-vladimirov: the `*/` should be on the next line

rosen-vladimirov:
I'm not sure projectSettings is the correct place to add the `clean` property, but I'm not sure where should we add it. One way is to change entirely the API of the method and make it correct (we pass strings, objects, more than three params, etc.), but this requires changes in Fusion as well.
I fully agree that adding it in the projectSettings is the fastest way to achieve the required functionality, so we can merge it and improve it in the future. 
Ping @Mitko-Kerezov , @TsvetanMilanov , @tsvetie

Mitko-Kerezov:
Regarding the questions:
> Can the client determine whether there's need for clean and full build?
    
If by __he client__ you're referring to the fusion then I guess it'll be a checkbox or sth.

> Would the --clean option be clear enough for the users? How they'll know...?

We've not yet devised a way of writing and displaying help for the commands introduced by extensions but we have a few ideas in mind.

> I'm not sure projectSettings is the correct place to add the clean property

There seems to be no other place to add the property so I'm guessing we stick with this implementation. I am reluctant to change the API method into accepting a single method at this point.

TsvetanMilanov:
@rosen-vladimirov I think it is OK to leave the clean property in the projectSettings. Maybe we should consider change in the API if we need to add more properties like clean to the projectSettings.

PR #22, opened 2017-06-01
Add runtime version in local package.json

nadyaA:
We need to know what runtime version is used for build as incremental builds
have no logic of detecting one.

### DIFF
--- a/lib/services/cloud-build-service.ts
+++ b/lib/services/cloud-build-service.ts
@@ -25,6 +25,7 @@ export class CloudBuildService extends EventEmitter implements ICloudBuildServic
 		private $cloudBuildOutputFilter: ICloudBuildOutputFilter,
 		private $mobileHelper: Mobile.IMobileHelper,
 		private $projectHelper: IProjectHelper,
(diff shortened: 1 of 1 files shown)


lib/services/cloud-build-service.ts
```diff
@@ -471,7 +476,7 @@ export class CloudBuildService extends EventEmitter implements ICloudBuildServic
 		if (!runtimeVersion && coreModulesVersion) {
 			// no runtime added. Let's find out which one we need based on the tns-core-modules.
 			if (semver.valid(coreModulesVersion)) {
-				runtimeVersion = `${semver.major(coreModulesVersion)}.${semver.minor(coreModulesVersion)}.*`;
+				runtimeVersion =  await this.getLatestMatchingVersion(runtimePackageName, `~${semver.major(coreModulesVersion)}.${semver.minor(coreModulesVersion)}.0`);
```
Mitko-Kerezov: Not a merge-stopper but you could extract a helper method `versionString => ~${semver.major(versionString)}.${semver.minor(versionString)}.0`

PR #23, opened 2017-06-01
Implement getCloudBuildOutputDirectory method

Mitko-Kerezov:
Used to get the path to the directory where the build output may be found.

Ping @nadyaA @rosen-vladimirov @TsvetanMilanov

### DIFF
--- a/lib/definitions/cloud-build-service.d.ts
+++ b/lib/definitions/cloud-build-service.d.ts
@@ -76,6 +76,13 @@ interface ICloudBuildService {
 		projectId: string,
 		androidBuildData?: IAndroidBuildData,
 		iOSBuildData?: IIOSBuildData): Promise<void>;
+
+	/**
+	 * Returns the path to the directory where the build output may be found.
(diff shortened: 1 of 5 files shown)


lib/definitions/cloud-build-service.d.ts
```diff
@@ -76,6 +76,13 @@ interface ICloudBuildService {
 		projectId: string,
 		androidBuildData?: IAndroidBuildData,
 		iOSBuildData?: IIOSBuildData): Promise<void>;
+
+	/**
+	 * Returns the path to the directory where the build output may be found.
+	 * @param {ICloudBuildOutputDirectoryOptions} options Options that are used to determine the build output directory.
+	 * @returns {string} The build output directory.
+	 */
+	getCloudBuildOutputDirectory(options: ICloudBuildOutputDirectoryOptions): string;
```
rosen-vladimirov: It is in the cloud build service, so we can rename it to `getBuildOutputDirectory`

PR #24, opened 2017-06-02
Move the login complete page in the server

TsvetanMilanov:
Since the logout complete page is in our server we should move the login complete page there too. This requires to get new parameter from the query string of the request to the localhost - loginCompleteUrl. Then we just redirect to it instead of serving local file.

`cloudBuildService` changes:
- change the fullOutput property to be the actual full output of the build.
- change the `output` event to `buildOutput`
- change the `buildOutput` event data from string to:
```JSON
{
    "pipe": "stdout",
    "data": "Add platform ios with runtime version 2.5.*"
}
```


lib/services/authentication-service.ts
```diff
@@ -32,19 +31,21 @@ export class AuthenticationService implements IAuthenticationService {
...
 		this.localhostServer = this.$httpServer.createServer({
 			routes: {
-				"/": async (request: ServerRequest, response: ServerResponse) => {
+				"/": (request: ServerRequest, response: ServerResponse) => {
 					this.$logger.debug("Login complete: " + request.url);
 					const parsedUrl = parse(request.url, true);
 					const loginResponse = parsedUrl.query.response;
 					if (loginResponse) {
-						await this.serveLoginFile("end.html")(request, response);
+						response.statusCode = 302;
```
Mitko-Kerezov: You can add this number to `constants.HTTP_STATUS_CODES`

PR #28, opened 2017-06-08
Use git to upload projects

TomaNikolov:
* Create credential helper to authenticate git in aws code commit. 
* We will use git to upload only the diff between the cloud and the local repository. This way we will speed up the time for the cloud builds.

### DIFF
--- a/resources/git/aws-credential-helper/credential-helper
+++ b/resources/git/aws-credential-helper/credential-helper
@@ -1,0 +1,6 @@
+#!/bin/sh
+basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
+
+node "$basedir/code-commit-credential-helper.js"
+
+exit
--- a/resources/git/aws-credential-helper/credential-helper
+++ b/resources/git/aws-credential-helper/credential-helper
@@ -1,4 +1,6 @@
 #!/bin/sh
+
+# Get the absolute path to the resources/git/aws-credential-helper folder.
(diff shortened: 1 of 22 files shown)


resources/git/template-.gitignore
```diff
@@ -0,0 +1,49 @@
+# .abignore lets you configure which of your files and folders should be excluded from your application package during the build process.
```
TsvetanMilanov: Do we need AppBuilder related comments?

resources/git/aws-credential-helper/credential-helper
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
+
+node "$basedir/code-commit-credential-helper.js"
```
TsvetanMilanov: Can you add comment with link to the description of this black magic?

server-configs/config-dev.json
```diff
@@ -9,6 +9,9 @@
 		},
 		"emulators-service": {
 			"subdomain": "dev-emulators"
+		},
+		"code-commit": {
```
TsvetanMilanov: You should add `-service` in all config.json files.

server-configs/config-production.json
```diff
@@ -2,16 +2,16 @@
 	"stage": "production",
 	"cloudServices": {
 		"auth-service": {
-			"subdomain": "authentication"
+			"subdomain": "authentication-service"
```
TsvetanMilanov: Remove the `-service` from the subdomain

PR #35, opened 2017-06-13
Add platform before building

Mitko-Kerezov:
This mimics the local build closely and may prevent some future problems.
The short-term benefit is the fact that by adding a platform the runtime's version is written in the package.json and is therefore respected in the cloud.

Ping @TsvetanMilanov @rosen-vladimirov

### DIFF
--- a/lib/services/cloud-build-service.ts
+++ b/lib/services/cloud-build-service.ts
@@ -56,16 +56,19 @@ export class CloudBuildService extends EventEmitter implements ICloudBuildServic
 			buildProps = await this.getiOSBuildProperties(projectSettings, buildProps, iOSBuildData);
 		}
 
+		const projectData = this.$projectDataService.getProjectData(projectSettings.projectDir);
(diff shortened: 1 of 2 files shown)


lib/services/cloud-build-service.ts
```diff
@@ -55,17 +56,18 @@ export class CloudBuildService extends EventEmitter implements ICloudBuildServic
 			buildProps = await this.getiOSBuildProperties(projectSettings, buildProps, iOSBuildData);
 		}
 
+		const projectData = this.$projectDataService.getProjectData(projectSettings.projectDir);
+		const platformAlreadyInstalled = _.includes(this.$platformService.getInstalledPlatforms(projectData).map(p => p.toLowerCase()), platform.toLowerCase());
```
TsvetanMilanov: You can change this line to:
```JavaScript
const platformAlreadyInstalled = _(this.$platformService.getInstalledPlatforms(projectData))
                                    .map(p => p.toLowerCase())
                                    .includes(platform.toLowerCase());
```

PR #37, opened 2017-06-16
Pass mobileProvision data to prepare

rosen-vladimirov:
Pass the full mobileProvision data to prepare method, so it will not execute `security` command which is available only on macOS.


rosen-vladimirov:
Merging as it has to be released ASAP. Will fix comments later and release a new version in case it is required.

Mitko-Kerezov:
Code looks okay to me 👍

PR #41, opened 2017-06-20
Get user avatar.

TomaNikolov:
* Introduce public method which return the url to the user's avatar.
* Fix typo `updloadToS3 -> uploadToS3`


lib/services/user-service.ts
```diff
@@ -68,6 +73,15 @@ export class UserService implements IUserService {
 		this.$fs.deleteFile(this.getUserFilePath());
 	}
 
+	public async getUserAvatar(): Promise<string> {
+		if (!this.hasUser()) {
+			return null;
+		}
+
+		const userInfo = await this.$authCloudService.getUserInfo();
+		return userInfo.externalAvatarUrl;
```
rosen-vladimirov: what about the `gravatar`?

rosen-vladimirov: Maybe we should return object with properties:
```JSON
{
    avatar_type
    gravatar
    externalAvatarUrl
}
```

TomaNikolov: It seems like the gravatar is not used and there is no UI to set it. Also, the avatarType is some magic int with the only known value of `2` which mean `External`

lib/definitions/user-service.d.ts
```diff
@@ -35,4 +35,10 @@ interface IUserService {
 	 * Removes the current user data.
 	 */
 	clearUserData(): void;
+
+	/**
+	 * Return the URL where the avatar picture can be downloaded from.
+	 * @returns {boolean} Return the URL where the avatar picture can be downloaded from. It will return null if the user does not have avatar or it is not logged in.
```
TsvetanMilanov: returns {Promise\<string\>}

PR #46, opened 2017-07-13
Check for unsupported git version on Windows.

TomaNikolov:
If the version of git is below 2.9.0 we shouldn't use git or windows credential manager will prompt for username and password.

### DIFF
--- a/lib/services/git-service.ts
+++ b/lib/services/git-service.ts
@@ -85,10 +85,13 @@ export class GitService implements IGitService {
 			// this resets the helper list to empty (so you may override a helper set by a lower-priority config file by configuring the empty-string helper,
 			// followed by whatever set of helpers you would like).
 			// https://git-scm.com/docs/gitcredentials
-			if (this.$hostInfo.isWindows &&
-				gitMajorVersion === GitService.MINIMAL_GIT_MAJOR_VERSION &&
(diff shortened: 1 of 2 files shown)


lib/services/git-service.ts
```diff
@@ -85,10 +85,13 @@ export class GitService implements IGitService {
...
-			if (this.$hostInfo.isWindows &&
-				gitMajorVersion === GitService.MINIMAL_GIT_MAJOR_VERSION &&
-				gitMinorVersion >= GitService.MINIMAL_GIT_MINOR_VERSION) {
-				await this.executeCommand(projectDir, ["config", "--local", `credential.${remoteUrl.httpRemoteUrl}.helper`, ""]);
+			if (this.$hostInfo.isWindows) {
+				if (gitMajorVersion === GitService.MINIMAL_GIT_MAJOR_VERSION &&
+					gitMinorVersion >= GitService.MINIMAL_GIT_MINOR_VERSION) {
+					await this.executeCommand(projectDir, ["config", "--local", `credential.${remoteUrl.httpRemoteUrl}.helper`, ""]);
+				} else {
+					throw new Error("Unsupported Git version. The minimal supported version is 2.9.0. Please update.");
```
rosen-vladimirov: Can you please add the current git version in the error message

PR #48, opened 2017-07-21
[WIP] Read cloud env from config files

TsvetanMilanov:
We can override some of the environment variables on our server by passing special headers. We should provide a way to set the env vars and their values in the headers of the request. The best place is in the server configuration files. The environment variables can be service specific or global.

Example:
```JSON
{
  "apiVersion": "v1",
  "domainName": "somedomain.com",
  "serverProto": "https",
  "stage": "production",
  "cloudServices": {
    "first-service": {
      "subdomain": "first",
      "cloudEnv": {
        "TEST": "TEST1"
      }
    },
    "second-service": {
      "subdomain": "second",
      "cloudEnv": {
        "TEST": "TEST2"
      }
    },
    "third-service": {
      "subdomain": "third"
    }
  },
  "cloudEnv": {
    "TEST": "TEST"
  }
}
```

### DIFF
--- a/lib/definitions/server-config-manager.d.ts
+++ b/lib/definitions/server-config-manager.d.ts
@@ -18,21 +18,21 @@ interface IServerConfigManager {
 	printConfigData(): void;
 }
 
-interface IServerConfig {
+interface IServerConfig extends IDynamicIndex {
 	apiVersion?: string;
 	domainName: string;
 	serverProto?: string;
 	stage?: string;
+	cloudEnv?: IStringDictionary;
 	cloudServices: IDictionary<ICloudServiceConfig>;
-	[index: string]: string | IDictionary<ICloudServiceConfig>;
 }
 
-interface ICloudServiceConfig {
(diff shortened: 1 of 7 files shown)


lib/definitions/server-config-manager.d.ts
```diff
@@ -23,16 +23,18 @@ interface IServerConfig {
 	domainName: string;
 	serverProto?: string;
 	stage?: string;
+	cloudEnv?: IStringDictionary;
 	cloudServices: IDictionary<ICloudServiceConfig>;
-	[index: string]: string | IDictionary<ICloudServiceConfig>;
+	[index: string]: any;
```
Mitko-Kerezov: You can extract this as a separate interface named something like `IDynamicIndex` and thus reuse it here and on the line below

TsvetanMilanov:
Closing this PR and keep the branch because the feature on the server will not be finished soon.

PR #52, opened 2017-08-24
Implement codesign cloud service and command

nadyaA:
There is new server logic for generation of iOS codesign files. With providing apple credentials for free apple account, user can generate development .p12 and .mobileprovision files to build their project in debug configuration. The implementation is similar to already existing one for cloud build. That is why some common logic is extracted. Naming of some of the items cannot be changed as it will require many follow-up changes that might be a topic for another PR. For consistency request properties are changed to be camel case as they're expected on the server.

http://teampulse.telerik.com/view#item/346464

https://github.com/Icenium/BuilderDaemon/pull/101

### DIFF
--- a/lib/commands/cloud-codesign.ts
+++ b/lib/commands/cloud-codesign.ts
@@ -1,7 +1,11 @@
-export class CloudCodesign implements ICommand {
+import { isInteractive } from "../helpers";
+
+export class CloudCodesignCommand implements ICommand {
+	private parametersValidationText = "The command has only 3 valid parameters - Apple account id, Apple account password and optional parameter clean for force re-invokation of certificate, if any.";
 	public allowedParameters: ICommandParameter[];
 
 	constructor(private $logger: ILogger,
+		private $errors: IErrors,
 		private $projectData: IProjectData,
 		private $prompter: IPrompter,
 		private $options: IOptions,
@@ -14,21 +18,33 @@ export class CloudCodesign implements ICommand {
 		let password = args[1];
 
 		if (!username) {
-			username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
+			if (isInteractive) {
+				username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
+			} else {
+				this.$errors.fail(this.parametersValidationText);
+			}
 		}
 
 		if (!password) {
-			password = await this.$prompter.getPassword("Apple ID password");
+			if (isInteractive) {
+				password = await this.$prompter.getPassword("Apple ID password");
+			} else {
+				this.$errors.fail(this.parametersValidationText);
+			}
 		}
 
-		this.$logger.info("Executing cloud codesign command");
+		this.$logger.info("Generating codesign files in the cloud");
 		const codesignData = { username, password, clean: this.$options.clean };
-		await this.$cloudCodesignService.generateCodesignFiles(codesignData, this.$projectData);
+		await this.$cloudCodesignService.generateCodesignFiles(codesignData, this.$projectData.projectDir);
 	}
 
 	public async canExecute(args: string[]): Promise<boolean> {
+		if (args.length > 3) {
+			this.$errors.fail(this.parametersValidationText);
+		}
+
 		return true;
 	}
 }
 
-$injector.registerCommand("codesign|cloud", CloudCodesign);
+$injector.registerCommand("codesign|cloud", CloudCodesignCommand);
--- a/lib/commands/cloud-codesign.ts
+++ b/lib/commands/cloud-codesign.ts
@@ -1,10 +1,15 @@
 import { isInteractive } from "../helpers";
 
 export class CloudCodesignCommand implements ICommand {
-	private parametersValidationText = "The command has only 3 valid parameters - Apple account id, Apple account password and optional parameter clean for force re-invokation of certificate, if any.";
+	private readonly parametersValidationText = "The command has only two valid parameters - Apple account id and Apple account password.";
+	// Currently only iOS codesign generation is supported.
+	private readonly platform = this.$devicePlatformsConstants.iOS;
+	private devices: Mobile.IDeviceInfo[];
 	public allowedParameters: ICommandParameter[];
 
 	constructor(private $logger: ILogger,
+		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
+		private $devicesService: Mobile.IDevicesService,
 		private $errors: IErrors,
 		private $projectData: IProjectData,
 		private $prompter: IPrompter,
@@ -18,29 +23,36 @@ export class CloudCodesignCommand implements ICommand {
 		let password = args[1];
 
 		if (!username) {
-			if (isInteractive) {
-				username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
-			} else {
-				this.$errors.fail(this.parametersValidationText);
-			}
+			username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
 		}
 
 		if (!password) {
-			if (isInteractive) {
-				password = await this.$prompter.getPassword("Apple ID password");
-			} else {
-				this.$errors.fail(this.parametersValidationText);
-			}
+			password = await this.$prompter.getPassword("Apple ID password");
 		}
 
 		this.$logger.info("Generating codesign files in the cloud");
-		const codesignData = { username, password, clean: this.$options.clean };
+
+		const codesignData = {
+			username, password,
+			clean: this.$options.clean,
+			platform: this.platform,
+			attachedDevices: this.devices
+		};
+
 		await this.$cloudCodesignService.generateCodesignFiles(codesignData, this.$projectData.projectDir);
 	}
 
 	public async canExecute(args: string[]): Promise<boolean> {
-		if (args.length > 3) {
+		if (args.length > 2 || (!isInteractive() && args.length < 2)) {
 			this.$errors.fail(this.parametersValidationText);
+		}
+
+		await this.$devicesService.detectCurrentlyAttachedDevices({ shouldReturnImmediateResult: false, platform: this.platform });
+		this.devices = this.$devicesService.getDeviceInstances()
+			.filter(d => !d.isEmulator && d.deviceInfo.platform.toLowerCase() === this.platform.toLowerCase())
+			.map(d => d.deviceInfo);
+		if (!this.devices || this.devices.length === 0) {
+			this.$errors.fail("Please attach iOS devices for which to generate codesign files.");
 		}
 
 		return true;
--- a/lib/commands/cloud-codesign.ts
+++ b/lib/commands/cloud-codesign.ts
@@ -1,0 +1,34 @@
+export class CloudCodesign implements ICommand {
+	public allowedParameters: ICommandParameter[];
+
+	constructor(private $logger: ILogger,
+		private $projectData: IProjectData,
+		private $prompter: IPrompter,
+		private $options: IOptions,
+		private $cloudCodesignService: ICloudCodesignService) {
+		this.$projectData.initializeProjectData();
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		let username = args[0];
+		let password = args[1];
+
+		if (!username) {
+			username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
+		}
+
+		if (!password) {
+			password = await this.$prompter.getPassword("Apple ID password");
+		}
+
+		this.$logger.info("Executing cloud codesign command");
+		const codesignData = { username, password, clean: this.$options.clean };
+		await this.$cloudCodesignService.generateCodesignFiles(codesignData, this.$projectData);
+	}
+
+	public async canExecute(args: string[]): Promise<boolean> {
+		return true;
+	}
+}
+
+$injector.registerCommand("codesign|cloud", CloudCodesign);
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -1,5 +1,5 @@
 export const CLOUD_TEMP_DIR_NAME = ".cloud";
-export const CODESIGN_TEMP_DIR_NAME = ".codesign";
+export const CODESIGN_FILES_DIR_NAME = "codesign_files";
 export const RELEASE_CONFIGURATION_NAME = "release";
 
 export const CRYPTO = {
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -1,4 +1,5 @@
 export const CLOUD_TEMP_DIR_NAME = ".cloud";
+export const CODESIGN_TEMP_DIR_NAME = ".codesign";
 export const RELEASE_CONFIGURATION_NAME = "release";
 
 export const CRYPTO = {
@@ -81,6 +82,7 @@ export const DISPOSITIONS = {
 	PACKAGE_GIT: "PackageGit",
 	BUILD_RESULT: "BuildResult",
 	PROVISION: "Provision",
+	CERTIFICATE: "Certificate",
 	KEYCHAIN: "Keychain",
 	CRYPTO_STORE: "CryptoStore"
 };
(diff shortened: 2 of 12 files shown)


lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,133 @@
...
+		const codesignRequest = await this.prepareCodesignRequest(buildId, codesignData, projectData);
+		const codesignResponse: IServerResponse = await this.$buildCloudService.generateCodesignFiles(codesignRequest);
+		this.$logger.trace(`Codesign response: ${JSON.stringify(codesignResponse)}`);
+
+		let codesignResult: IServerResult;
+		try {
+			await this.waitForServerOperationToFinish(buildId, codesignResponse);
+		} catch (ex) {
+			this.$logger.trace("Codesign generation failed with err: ", ex);
+		} finally {
```
rosen-vladimirov: what's the purpose of this `finally` - we catch the error, so you can safely remove the finally block and directly set:
```TypeScript
const codesignResult = await this.getObjectFromS3File<IServerResult>(codesignResponse.resultUrl);
```

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,133 @@
...
+		if (!codesignResult.buildItems || !codesignResult.buildItems.length) {
+			// Something failed.
+			this.$errors.failWithoutHelp(`Codesign failed. Reason is: ${codesignResult.errors}. Additional information: ${codesignResult.stderr}.`);
+		}
+
+		this.$logger.info(`Finished ${codesignInformationString} successfully. Downloading result...`);
+
+		const localCodesignResults = await this.downloadServerResults(codesignResult, {
+			projectDir: projectData.projectDir,
+			platform: this.platform,
```
rosen-vladimirov: In case we'll use the same class for Android generation in the future, you should not use the platform from `this`, but receive it from the caller.

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	private async prepareCodesignRequest(buildId: string,
+		codesignData: ICodesignData,
+		projectData: IProjectData): Promise<any> {
+
+		const sanitizedProjectName = this.$projectHelper.sanitizeName(projectData.projectName);
+		await this.$devicesService.detectCurrentlyAttachedDevices({ shouldReturnImmediateResult: false, platform: this.platform });
+		let attachedDevices = this.$devicesService.getDeviceInstances();
+		attachedDevices = attachedDevices.filter(d => d.deviceInfo.platform.toLowerCase() === this.platform.toLowerCase());
```
rosen-vladimirov: so we'll included the currently running simulators in the provision?

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,133 @@
...
+		const devices = attachedDevices.map(d => ({ displayName: d.deviceInfo.displayName, identifier: d.deviceInfo.identifier }));
+
+		return {
+			buildId,
+			appId: projectData.projectId,
+			appName: sanitizedProjectName,
+			clean: codesignData.clean,
+			username: codesignData.username,
+			password: codesignData.password,
+			devices: devices
```
rosen-vladimirov: just `devices`

lib/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const CLOUD_TEMP_DIR_NAME = ".cloud";
+export const CODESIGN_TEMP_DIR_NAME = ".codesign";
```
rosen-vladimirov: This does not seem like temp dir name, i.e. we download the certificate and provision there, it is inside the project and there's nothing temporarily in it.
Also why the name start with `.` - do we want to download the provision and certificate and leave the user wondering where are they?

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,34 @@
...
+		private $cloudCodesignService: ICloudCodesignService) {
+		this.$projectData.initializeProjectData();
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		let username = args[0];
+		let password = args[1];
+
+		if (!username) {
+			username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
```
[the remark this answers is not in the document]
nadyaA: We had a discussion about the parameters approach with @Mitko-Kerezov and @TsvetanMilanov . The idea here is to have at least some prompts as you do not have explicit --options and you do not have help. This way you do not have to know the exact order of the parameters and all of them as a matter of fact.  Similar implementation may be found here - https://github.com/NativeScript/nativescript-cli/blob/master/lib/commands/appstore-upload.ts#L38 , so users are familiar with this behavior.

I know this is also meant to be used for automation scenarios, but in such cases, I think it is highly possible the parameters are passed correctly. If not it will be one time debug to find that you are missing parameters. 
Of course, I am open to further discussions, may be @Mitko-Kerezov and @TsvetanMilanov  might share opinion?

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+		if (!password) {
+			password = await this.$prompter.getPassword("Apple ID password");
```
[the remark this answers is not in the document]
nadyaA: same from me as well - why noone - there is someone that invoked the command 😸

lib/commands/cloud-codesign.ts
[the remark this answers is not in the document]
rosen-vladimirov: `I know this is also meant to be used for automation scenarios, but in such cases, I think it is highly possible the parameters are passed correctly.`

**highly possible** does not mean that they are passed correctly. 
`it will be one time debug to find that you are missing parameters.` - do you expect users to debug CLI ?

The idea is to use `isInteractive()` method and in case not and in case all required arguments are not passed, to fail with descriptive message instead of hanging.
You can handle this in `canExecute` method of the command.

lib/commands/cloud-codesign.ts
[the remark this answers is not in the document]
rosen-vladimirov: So, he'll be able to answer you?
image

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,133 @@
...
+		super($fs, $httpClient, $logger);
+		this.failedToStartError = "Failed to start cloud codesign generation.";
+		this.failedError = "Codesign generation failed.";
+		this.operationInProgressStatus = "Working";
+	}
+
+	public async generateCodesignFiles(codesignData: ICodesignData,
+		projectData: IProjectData): Promise<ICodesignResultData> {
+		codesignData.clean = codesignData.clean === undefined ? true : codesignData.clean;
+		const buildId = uuid.v4();
```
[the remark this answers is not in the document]
nadyaA: For consistency - I find it confusing to have just here the requestId and everywhere else buildId as it refers to same parameter.  I explained the naming [here](https://github.com/NativeScript/nativescript-cloud/pull/52/files/44396757a1f7ad7e8ff3c65c9e7fa268c89d9e69#diff-ce297c9683b251e344171ba2035b5ba6R6). As I reviewed the code, I thought having two is confusing.

lib/services/cloud-codesign-service.ts
nadyaA: The idea here is that unlike the build lambda, the codesign one is using only two files to share what happened. I.e. the resultsUrl consists of all input and result items if available. We catch error that is thrown in the try, but if there is some error in the build machine scripts it will be written in the results file. So we always want to read the file. Otherwise we do not see why the build machine script failed. For example providing wrong credentials will come as stderr and will be logged in the results file. [Here](https://github.com/NativeScript/nativescript-cloud/pull/52/files/44396757a1f7ad7e8ff3c65c9e7fa268c89d9e69#diff-5cdce4f47c4d43f2b52f9f9b4f256d1dR83) you can see that we need the codesignResult in any case.

lib/services/cloud-codesign-service.ts
nadyaA: Of course. It will not be hard-coded as well. Currently I added it for consistency - having the folders structure with iOS in it is neat for iOS codesign generation. I just laid some foundations in case we'll have other platforms. Further refactoring will be needed at such case.

lib/constants.ts
nadyaA: I wasn't sure - do we make them wonder where are the binaries? I supposed there should be some additional work for saving/using these. Wasn't sure. Anyway will remove the hidden name and temp identifier.

lib/services/cloud-codesign-service.ts
rosen-vladimirov: I understand this, but I still do not see why you need `finally`, you can just use:
```TypeScript
try {
} catch (err) 
...
}

const codesignResult = await this.getObjectFromS3File<IServerResult>(codesignResponse.resultUrl);
```

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,50 @@
...
+			}
+		}
+
+		this.$logger.info("Generating codesign files in the cloud");
+		const codesignData = { username, password, clean: this.$options.clean };
+		await this.$cloudCodesignService.generateCodesignFiles(codesignData, this.$projectData.projectDir);
+	}
+
+	public async canExecute(args: string[]): Promise<boolean> {
+		if (args.length > 3) {
```
rosen-vladimirov: shouldn't this be:
```TypeScript
if (args.length > 2) {
```
Currently we allow the command: `tns codesign cloud AppleId ApplePassword pesho`

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,50 @@
+import { isInteractive } from "../helpers";
+
+export class CloudCodesignCommand implements ICommand {
+	private parametersValidationText = "The command has only 3 valid parameters - Apple account id, Apple account password and optional parameter clean for force re-invokation of certificate, if any.";
```
rosen-vladimirov: in fact they are two, not 3.  `--` parameters are not included in `args`

rosen-vladimirov: Also please use the word `two` instead of 2

lib/definitions/cloud-codesign-service.d.ts
```diff
@@ -0,0 +1,33 @@
...
+	outputFilesPaths: string[];
+}
+
+/**
+ *  Describes specific data required for codesigning files to be generated.
+ */
+interface ICodesignData extends ICredentials {
+	/**
+	 * Whether to revoke and reissue certificate even if it is not expired. This option is necessary as re-downloading the .p12 file is impossible.
+	 * It provides consumers with opportunity to deliberately reissue account's development certificate. It will default to true as it is the excpected behaviour.
```
rosen-vladimirov: `excpected` - `expected`

lib/definitions/cloud-service.d.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Describes the result from the server operation.
+ */
+interface IServerResult {
+	errors: string[];
+	code: number;
+	stdout: string;
+	stderr: string;
+	buildItems: IServerItem[]; // naming is due to initial implementation that is related to other repos.
```
rosen-vladimirov: JS docs instead of comment ?

lib/services/cloud-build-service.ts
```diff
@@ -5,35 +5,41 @@ import * as forge from "node-forge";
...
-	private static BUILD_COMPLETE_STATUS = "Success";
-	private static BUILD_IN_PROGRESS_STATUS = "Building";
-	private static BUILD_FAILED_STATUS = "Failed";
+export class CloudBuildService extends CloudService implements ICloudBuildService {
 	private static DEFAULT_VERSION = "3.0.0";
 
-	constructor(private $buildCloudService: IBuildCloudService,
+	protected get failedError() {
+		return "Build failed.";
+	};
```
rosen-vladimirov: new line after the `};`

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,152 @@
...
+import { CloudService } from "./cloud-service";
+
+export class CloudCodesignService extends CloudService implements ICloudCodesignService {
+	private get platform() {
+		return this.$devicePlatformsConstants.iOS;
+	};
+
+	protected get failedError() {
+		return "Generation of codesign files failed.";
+	};
```
rosen-vladimirov: new lines after `};` please

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,50 @@
...
+		private $cloudCodesignService: ICloudCodesignService) {
+		this.$projectData.initializeProjectData();
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		let username = args[0];
+		let password = args[1];
+
+		if (!username) {
+			if (isInteractive) {
```
rosen-vladimirov: You can check for `isInteractive` in `canExecute` method, as you are currently failing with the same error. For example:
```TypeScript
public async canExecute(args: string[]): Promise<boolean> {
	if (args.length > 3) {
		this.$errors.fail(this.parametersValidationText);
	}

	if (!isInteractive() && args.length < 2) {
		this.$errors.fail(this.parametersValidationText);
	}
	
	return true;
}
```

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,50 @@
...
+				username = await this.$prompter.getString("Apple ID", { allowEmpty: false });
+			} else {
+				this.$errors.fail(this.parametersValidationText);
+			}
+		}
+
+		if (!password) {
+			if (isInteractive) {
```
rosen-vladimirov: isInteractive is method, you should call it

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,152 @@
...
+		// no specific implementation needed.
+	}
+
+	private async prepareCodesignRequest(buildId: string,
+		codesignData: ICodesignData,
+		projectData: IProjectData): Promise<any> {
+
+		const sanitizedProjectName = this.$projectHelper.sanitizeName(projectData.projectName);
+		await this.$devicesService.detectCurrentlyAttachedDevices({ shouldReturnImmediateResult: false, platform: this.platform });
+		let attachedDevices = this.$devicesService.getDeviceInstances();
```
rosen-vladimirov:
```
const attachedDevices = this.$devicesService.getDeviceInstances()
				filter(d => !d.isEmulator && d.deviceInfo.platform.toLowerCase() === this.platform.toLowerCase());
```

lib/services/cloud-codesign-service.ts
```diff
@@ -0,0 +1,141 @@
...
+		return path.join(options.projectDir, constants.CODESIGN_FILES_DIR_NAME, options.platform.toLowerCase());
+	}
+
+	private validateParameteres(codesignData: ICodesignData,
+		projectDir: string) {
+		if (!codesignData || !codesignData.username || !codesignData.password) {
+			this.$errors.failWithoutHelp(`Codesign failed. Reason is missing code sign data. Apple Id and Apple Id password are required..`);
+		}
+
+		if (!projectDir || projectDir === "") {
```
rosen-vladimirov:
```TypeScript
if (!projectDir) {
}
```
is enough ;)

lib/commands/cloud-codesign.ts
```diff
@@ -0,0 +1,71 @@
...
+				password = await this.$prompter.getPassword("Apple ID password");
+			} else {
+				this.$errors.fail(this.parametersValidationText);
+			}
+		}
+
+		this.$logger.info("Generating codesign files in the cloud");
+
+		const attachedDevices: Mobile.ISimpleDevice[] = this.devices.map(d => ({ displayName: d.deviceInfo.displayName, identifier: d.deviceInfo.identifier }));
```
rosen-vladimirov: can't we make the attachedDevices just instance of `Mobile.IDeviceInfo` and map should be: `this.devices.map (d => d.deviceInfo);`
Also where is this `ISimpleDevice`, I couldn't find it

PR #61, opened 2017-10-09
Add accounts

TsvetanMilanov:
Some of the cloud functionality will become paid. This requires the account abstraction.
Currently we have users and we use them in our cloud services. Now we are going to use the users as part of accounts. Each user has personal account. Also the user A can add to his/hers account the user B. This way the user B will be able to  use the subscriptions of the user A. 

This PR adds the following commands:
- `tns account` (`tns account list`) -> this command show all accounts of the current user.
- `tns account usage` -> this command show the remaining usage for each feature from the subscription of the provided account.

Added services to the public API:
- `accountsService`

### Breaking changes:
- `cloudBuildService.build` requires accountId.

### DIFF
--- a/lib/commands/account/usage.ts
+++ b/lib/commands/account/usage.ts
@@ -6,16 +6,16 @@ export class UsageCommand extends AccountCommandBase implements ICommand {
 	public allowedParameters: ICommandParameter[] = [];
 
 	public get dashedOptions() {
-		return this.$cloudOptionsProvider.dashedOptions;
+		return this.$nsCloudOptionsProvider.dashedOptions;
 	}
 
 	constructor($errors: IErrors,
-		$userService: IUserService,
+		$nsCloudUserService: IUserService,
 		private $nsCloudAccountsService: IAccountsCloudService,
-		private $cloudOptionsProvider: ICloudOptionsProvider,
+		private $nsCloudOptionsProvider: ICloudOptionsProvider,
 		private $options: ICloudOptions,
 		private $logger: ILogger) {
-		super($errors, $userService);
+		super($errors, $nsCloudUserService);
 	}
 
 	public async execute(args: string[]): Promise<void> {
--- a/lib/commands/account/usage.ts
+++ b/lib/commands/account/usage.ts
@@ -1,6 +1,6 @@
 import { AccountCommandBase } from "./account-command-base";
 import { UNLIMITED } from "../../constants";
-import { createTable } from "../../helpers";
+import { createTable, stringifyWithIndentation } from "../../helpers";
 
 export class UsageCommand extends AccountCommandBase implements ICommand {
 	public allowedParameters: ICommandParameter[] = [];
@@ -20,19 +20,27 @@ export class UsageCommand extends AccountCommandBase implements ICommand {
 
 	public async execute(args: string[]): Promise<void> {
 		const usage = await this.$nsCloudAccountsService.getUsageInfo(this.$options.accountId);
-		const table = createTable(["Feature", "Remaining usage"], usage.map(u => {
-			const result = [u.feature];
-			if (u.unlimited) {
-				result.push(UNLIMITED);
-			} else {
-				let remainingUsage = u.allowedUsage - u.usage;
-				result.push(remainingUsage.toString());
-			}
+		let output: string;
 
-			return result;
-		}));
+		if (this.$options.json) {
+			output = stringifyWithIndentation(usage);
+		} else {
+			const table = createTable(["Feature", "Remaining usage"], usage.map(u => {
+				const result = [u.feature];
+				if (u.unlimited) {
+					result.push(UNLIMITED);
+				} else {
+					const remainingUsage = u.allowedUsage - u.usage;
+					result.push(remainingUsage.toString());
+				}
 
-		this.$logger.out(table.toString());
+				return result;
+			}));
+
+			output = table.toString();
+		}
+
+		this.$logger.out(output);
 	}
 }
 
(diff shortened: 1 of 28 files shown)


lib/services/accounts-service.ts
```diff
@@ -0,0 +1,39 @@
...
+	public async getUsageInfo(accountIdOption: string): Promise<IUsageInfo[]> {
+		const account = await this.getAccountFromOption(accountIdOption);
+		return this.$nsCloudAccountsCloudService.getUsageInfo(account.id);
+	}
+
+	public async getAccountFromOption(accountIdOption: string): Promise<IAccount> {
+		const accounts = await this.getMyAccounts();
+		if (!accountIdOption) {
+			this.$errors.failWithoutHelp("Please provide accountId.");
+		} else {
```
rosen-vladimirov: No need of else here

lib/commands/account/usage.ts
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		const usage = await this.$nsCloudAccountsService.getUsageInfo(this.$options.accountId);
+		const table = createTable(["Feature", "Remaining usage"], usage.map(u => {
+			const result = [u.feature];
+			if (u.unlimited) {
+				result.push(UNLIMITED);
+			} else {
+				let remainingUsage = u.allowedUsage - u.usage;
```
rosen-vladimirov: const?

lib/services/accounts-service.ts
```diff
@@ -0,0 +1,39 @@
...
+			let selectedAccount = _.find(accounts, a => a.id === accountIdOption);
+			if (selectedAccount) {
+				return selectedAccount;
+			} else {
+				try {
+					const accountIndex = Number.parseInt(accountIdOption);
+					selectedAccount = accounts[accountIndex - 1];
+					if (selectedAccount) {
+						return selectedAccount;
+					} else {
```
rosen-vladimirov: no need of else here

lib/services/accounts-service.ts
```diff
@@ -0,0 +1,39 @@
...
+						this.$errors.failWithoutHelp("Invalid accountId index provided.");
+					}
+				} catch (err) {
+					this.$errors.failWithoutHelp("Invalid accountId option provided.");
```
rosen-vladimirov: `logger.trace` the real error

package.json
```diff
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "aws4": "1.6.0",
+    "cli-table": "https://github.com/telerik/cli-table/tarball/v0.3.1.2",
```
rosen-vladimirov: 😿

lib/commands/account/usage.ts
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		const usage = await this.$nsCloudAccountsService.getUsageInfo(this.$options.accountId);
+		let output: string;
+
+		if (this.$options.json) {
+			output = stringifyWithIndentation(usage);
+		} else {
+			const table = createTable(["Feature", "Remaining usage"], usage.map(u => {
```
Mitko-Kerezov: Since the command is named `usage` I'd expect the full usage to be listed - builds done as well as builds remaining.
So I'd suggest `Feature`, `Builds performed` and `Builds remaining` for example.

Note that this would impact the `json` output you provide also.

lib/definitions/cloud-services/accounts-cloud-service.d.ts
```diff
@@ -0,0 +1,54 @@
+interface IAccountsCloudService {
+	getAccounts(): Promise<IAccount[]>
+	getUsageInfo(accountId: string): Promise<IUsageInfo[]>;
```
Mitko-Kerezov: This could be extracted into a separate interface (for example `IGetUsageInfo` or `IAccountsServiceBase`) and be `extend`-ed here and in `IAccountsService`

lib/definitions/cloud-services/accounts-cloud-service.d.ts
```diff
@@ -0,0 +1,54 @@
+interface IAccountsCloudService {
+	getAccounts(): Promise<IAccount[]>
```
Mitko-Kerezov: Some docstrings would be nice

lib/definitions/acccounts-service.d.ts
```diff
@@ -0,0 +1,5 @@
+interface IAccountsService {
+	getMyAccounts(): Promise<IAccount[]>
```
Mitko-Kerezov: Docstrings would be great

lib/helpers.ts
```diff
@@ -1,7 +1,23 @@
...
+	const table = new Table({
+		head: headers,
+		chars: { "mid": "", "left-mid": "", "mid-mid": "", "right-mid": "" }
+	});
+
+	_.forEach(data, row => table.push(row));
+	return table;
+}
+
+export function stringifyWithIndentation(data: any, indentation?: string): string {
```
Mitko-Kerezov: Definitely not a merge-stopper but the `indentation` argument's actual type should be `string | number`

server-configs/config-local.json
```diff
@@ -11,8 +11,11 @@
 		"emulators-service": {
 			"fullHostName": "localhost:8002"
 		},
-		"code-commit-service": {
```
Mitko-Kerezov: Lost in translation?

PR #65, opened 2017-10-30
Add EULA service and check if EULA is accepted  

rosen-vladimirov:
Add `nsCloudEulaService` and check if EULA is accepted before each cloud call. The EULA must be accepted before executing any command, even before login. Users should not be able to access cloud infrastructure without accepting the EULA.

The EULA itself is located on `nativescript.org` - download it from there once per 24 hours and check if user had accepted it. In case it has not, do not allow them to access the cloud.

In `user-settings.json` there's a property called `acceptedEulaHash` - it contains the hashsum of the EULA at the moment when the user accepted it.

Rules for checking if user should accept EULA:
1. Check `user-settings.json` - in case there's no property `acceptedEulaHash` - users must accept EULA
2. In case there's `acceptedEulaHash`, download the EULA from `nativescript.org`.
2.1 In case the download it successful, compare the shasums of the downloaded file with the value of the `acceptedEulaHash`. In case they are different - EULA must be accepted, as it is different from the one the user had accepted before. In case they are equal, user should not accept anything and can continue with the work.
2.2 In case the download fail, allow the user to use the tooling - he has already accepted an EULA, which is most probably the EULA from our site.

There are several optimizations:
- EULA is downloaded only once per process.
- There are two public methods in eula-service - one of them will download the EULA only in case it has not been downloaded in the last 24 hours. The other one will force the download. It's up to the caller to decide which method to use.

### Add prompter for accepting EULA on each command
Add `eula-command-helper` that has two methods - one to accept EULA and one to ensure EULA is accepted.
Use the second method in each command that requires cloud operations and in case user had not accepted EULA - prompt him to accept it. In case the terminal is not interactive or the user does not accept the EULA, command will not be executed.

Add new command - `tns accept eula` - it should be used for automation before executing any other cloud related command. It will not show prompt, it just informs the user that EULA will be accepted and saves the SHA of the current EULA.

### Add lockfile for download and for checking current EULA mtime
In case multiple processes try to download the EULA, they should not do it simultaneously as it may cause unexpected failures or incorrect shasum checks.
In order to resolve the issue add locking of the actions by using lockfile on two places - first one is before the check for 24 hours, the other one is on the real download step.

### Add unit tests
Add unit tests for eula-service.


lib/services/eula-service.ts
```diff
@@ -0,0 +1,162 @@
...
+			this.$logger.trace("Checking EULA state: no local copy of EULA found - as user had already accepted previous version of the EULA, consider it as the current one, so no need to accept new.");
+			return false;
+		}
+
+		this.$logger.trace(`Checking EULA state: acceptedEulaHash is ${acceptedEulaHash}, currentEulaHash is ${currentEulaHash}.`);
+		return acceptedEulaHash !== currentEulaHash;
+	}
+
+	private async downloadLatestEula(opts: { shouldThrowError: boolean } = { shouldThrowError: false }): Promise<void> {
+		// TODO: Lock this action as multiple processes may end up doing this simultaneously
```
rosen-vladimirov: Remove this TODO - already implemented

lib/services/eula-service.ts
```diff
@@ -0,0 +1,162 @@
...
+				pipeTo: this.$fs.createWriteStream(this.pathToEula),
+				timeout: EulaConstants.timeout
+			});
+
+			this.$logger.trace(`Successfully downloaded EULA to ${this.pathToEula}.`);
+
+			this.isEulaDownloadedInCurrentProcess = true;
+		} catch (err) {
+			if (opts.shouldThrowError) {
+				this.$logger.trace("Unable to download latest EULA, but will rethrow the error as requested by called. Error is:", err);
```
rosen-vladimirov: `by caller`

lib/services/eula-service.ts
```diff
@@ -0,0 +1,162 @@
...
+
+		const lockFilePath = this.getLockFilePath("eula.check_download.lock");
+		try {
+			await this.$lockfile.lock(lockFilePath, this.getLockFileParams());
+			// download file only in case it has not been modified in the last 24 hours
+			const eulaFstat = this.$fs.exists(this.pathToEula) && this.$fs.getFsStats(this.pathToEula);
+			const timeToCheck = 24 * 60 * 60 * 1000;
+			const currentTime = this.$nsCloudDateTimeService.getCurrentEpochTime();
+
+			// TODO: Check if mtime is the correct fstat data.
```
rosen-vladimirov: Remove this TODO - already checked

test/services/eula-service.ts
```diff
@@ -0,0 +1,200 @@
...
+					assert.notInclude(loggedMessage, msg);
+				}
+			});
+
+		});
+
+	});
+
+	describe("acceptEula", () => {
+		// should fail in case unable to download eula
```
rosen-vladimirov: remove these comments, already added these tests

PR #69, opened 2017-11-03
Download EULA to temp file

rosen-vladimirov:
The EULA should be downloaded to a temp file as the `pipeTo` option of httpClient.httpRequest` immediately creates the file or overwrites it.
In case the download fails, we'll end up with 0KB EULA, which will trigger many incorrect behaviors on our side.
In order to fix this - move the EULA to its location after full download is finished.

### DIFF
--- a/lib/services/eula-service.ts
+++ b/lib/services/eula-service.ts
@@ -1,5 +1,6 @@
 import * as path from "path";
 import { EulaConstants } from "../constants";
+import * as temp from "temp";
 
 export class EulaService implements IEulaService {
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
@Mitko-Kerezov I agree, but this will require CLI release, so I'll keep it this way for the moment.

rosen-vladimirov:
@Mitko-Kerezov I've also remembered a problem we had in the past with `mv` - some Node.js methods fail in case you are trying to move a file from one share to another. In the past we had to replace some mv code with cp

PR #74, opened 2017-11-16
 Check if cloud remote is correct

TsvetanMilanov:
We need to check if the cloud remote is set correctly before we execute
our git operations. Currently if the remote is set to incorrect url we
will try to add it again and this will fail. We should change the wrong
remote url if the remote is added and if the remote url is incorrect.

### DIFF
--- a/lib/services/git-service.ts
+++ b/lib/services/git-service.ts
@@ -33,6 +33,16 @@ export class GitService implements IGitService {
 			await this.gitInit(projectSettings);
 		}
 
+		const isRemoteAdded = await this.gitCheckIfRemoteIsAdded(projectSettings, GitService.REMOTE_NAME);
+		if (isRemoteAdded) {
+			const isGitRemoteCorrect = await this.isGitRemoteSetToCorrectUrl(projectSettings, remoteUrl);
+			if (!isGitRemoteCorrect) {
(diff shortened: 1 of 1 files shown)


lib/services/git-service.ts
```diff
@@ -89,77 +95,91 @@ export class GitService implements IGitService {
...
+	private async gitSetRemoteUrl(projectSettings: INSCloudProjectSettings, remoteName: string, remoteUrl: IRemoteUrl) {
+		await this.executeCommand(projectSettings, ["remote", "set-url", remoteName, remoteUrl.httpRemoteUrl]);
+	}
+
+	private async gitCheckIfRemoteIsAdded(projectSettings: INSCloudProjectSettings, remoteName: string) {
+		try {
+			await this.executeCommand(projectSettings, ["remote", "get-url", remoteName]);
+			return true;
+		} catch (err) {
+			this.$logger.trace("Error while checking if remote is added: ", err);
```
TomaNikolov: Do we need to log this as an error since this is the expected behavior if the remote is not set yet?

TsvetanMilanov: This error is only in the debug logs. I decided to log it because it can be other error and we will be able to see only from the debug logs why this method returns false instead of true.

PR #78, opened 2017-11-29
Throw meaningful error if apple service is unavailable.

nadyaA:
Reported in this issue - https://github.com/NativeScript/sidekick-feedback/issues/92

It looks like apple is having sporadic maintenance on the developer site and some of our users are experiencing strange 403 Forbidden error. Seem like fastline are having similar issue as well - https://github.com/fastlane/fastlane/issues/10323

### DIFF
--- a/lib/services/cloud-codesign-service.ts
+++ b/lib/services/cloud-codesign-service.ts
@@ -56,7 +56,7 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
 	}
 
 	private validateParameteres(codesignData: ICodesignData,
-		projectDir: string) : void {
+		projectDir: string): void {
 		if (!codesignData || !codesignData.username || !codesignData.password) {
 			this.$errors.failWithoutHelp(`Codesign failed. Reason is missing code sign data. Apple Id and Apple Id password are required..`);
 		}
@@ -89,7 +89,13 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
 
 		if (!codesignResult.buildItems || !codesignResult.buildItems.length) {
 			// Something failed.
-			const err: any = new Error(`Codesign failed. Reason is: ${codesignResult.errors}.`);
+			let errText = codesignResult.errors || "";
+			if (errText.indexOf("403 Forbidden") > -1) {
+				this.$logger.trace(`Codesign errors: ${errText}`);
+				errText = "The Code Signing Assistance service is temporary unavailable. Please try again later.";
+			}
+
+			const err = <IStdError>new Error(`Codesign failed. Reason is: ${errText}.`);
 			err.stderr = codesignResult.stderr;
 			throw err;
 		}


lib/services/cloud-codesign-service.ts
```diff
@@ -89,7 +89,12 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
 
 		if (!codesignResult.buildItems || !codesignResult.buildItems.length) {
 			// Something failed.
-			const err: any = new Error(`Codesign failed. Reason is: ${codesignResult.errors}.`);
+			let errText = codesignResult.errors;
```
rosen-vladimirov: what's the type of `errors`, if it can be `null` or `undefined`, you should guard against null reference exception:
```TypeScript
let errText = codesignResult.errors || "";
```

lib/services/cloud-codesign-service.ts
```diff
@@ -89,7 +89,12 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
...
+			if (codesignResult.errors.indexOf("403 Forbidden") > -1) {
+				errText = "The Code Signing Assistance service is temporary unavailable. Please try again later"
```
rosen-vladimirov: maybe we should log the real `errors`:
```TypeScript
this.$logger.trace(`Codesign errors: ${errText}`);
```

lib/services/cloud-codesign-service.ts
```diff
@@ -89,7 +89,12 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
 
 		if (!codesignResult.buildItems || !codesignResult.buildItems.length) {
 			// Something failed.
-			const err: any = new Error(`Codesign failed. Reason is: ${codesignResult.errors}.`);
+			let errText = codesignResult.errors;
+			if (codesignResult.errors.indexOf("403 Forbidden") > -1) {
```
rosen-vladimirov: you can use `errText` here

lib/services/cloud-codesign-service.ts
```diff
@@ -89,7 +89,12 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
...
+				errText = "The Code Signing Assistance service is temporary unavailable. Please try again later"
+			}
+
+			const err: any = new Error(`Codesign failed. Reason is: ${errText}.`);
```
rosen-vladimirov: You can introduce a new type for this error and add the `stderr` in it. This way you'll not have to cast to any.

nadyaA:
https://github.com/telerik/mobile-cli-lib/pull/1042

lib/services/cloud-codesign-service.ts
```diff
@@ -89,7 +89,13 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
 
 		if (!codesignResult.buildItems || !codesignResult.buildItems.length) {
 			// Something failed.
-			const err: any = new Error(`Codesign failed. Reason is: ${codesignResult.errors}.`);
+			let errText = codesignResult.errors || "";
+			if (errText.indexOf("403 Forbidden") > -1) {
+				errText = "The Code Signing Assistance service is temporary unavailable. Please try again later";
```
rosen-vladimirov: . at the end

lib/services/cloud-codesign-service.ts
```diff
@@ -89,7 +89,13 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
...
+				this.$logger.trace(`Codesign errors: ${errText}`);
```
rosen-vladimirov: this line should be before resetting errText

PR #84, opened 2018-01-03
Add command flags to change global server config

TsvetanMilanov:
Each server PR build produces test dns record with api version in the
following format -> `pr<pr-number>`. To make the development cycle easier
we can add --apiVersion flag to the config apply/set commands which will
override the default apiVersion of the provided configuration. This way
when we want to use a specific server PR we can just execute `tns config
apply test --apiVersion pr<pr-number>`.

### DIFF
--- a/lib/commands/config/config-apply.ts
+++ b/lib/commands/config/config-apply.ts
@@ -1,12 +1,19 @@
 export class ConfigApplyCommand implements ICommand {
-	constructor(private $nsCloudServerConfigManager: IServerConfigManager,
-		private $stringParameterBuilder: IStringParameterBuilder) { }
-
 	public allowedParameters: ICommandParameter[] = [this.$stringParameterBuilder.createMandatoryParameter("Configuration name cannot be empty.")];
+
+	public get dashedOptions() {
+		return this.$nsCloudOptionsProvider.dashedOptions;
+	}
+
+	constructor(private $nsCloudOptionsProvider: ICloudOptionsProvider,
+		private $nsCloudServerConfigManager: IServerConfigManager,
+		private $options: ICloudOptions,
+		private $stringParameterBuilder: IStringParameterBuilder) { }
 
 	public async execute(args: string[]): Promise<void> {
 		const configurationName = args[0];
-		this.$nsCloudServerConfigManager.applyConfig(configurationName);
+		const globalServerConfig = { apiVersion: this.$options.apiVersion, serverProto: this.$options.serverProto };
+		this.$nsCloudServerConfigManager.applyConfig(configurationName, null, globalServerConfig);
 		this.$nsCloudServerConfigManager.printConfigData();
 	}
 }
--- a/lib/commands/config/config-apply.ts
+++ b/lib/commands/config/config-apply.ts
@@ -12,7 +12,7 @@ export class ConfigApplyCommand implements ICommand {
 
 	public async execute(args: string[]): Promise<void> {
 		const configurationName = args[0];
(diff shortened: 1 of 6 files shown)


lib/commands/config/config-apply.ts
```diff
@@ -1,12 +1,19 @@
...
+
+	constructor(private $nsCloudOptionsProvider: ICloudOptionsProvider,
+		private $nsCloudServerConfigManager: IServerConfigManager,
+		private $options: ICloudOptions,
+		private $stringParameterBuilder: IStringParameterBuilder) { }
+
 	public async execute(args: string[]): Promise<void> {
 		const configurationName = args[0];
-		this.$nsCloudServerConfigManager.applyConfig(configurationName);
+		const globalServerConfig = { apiVersion: this.$options.apiVersion, serverProto: this.$options.serverProto };
```
Mitko-Kerezov: As a personal preference and definitely not a merge-stopper, I'd suggest marking this as `IServerConfigBase` so that it becomes:
```TypeScript
const globalServerConfig: IServerConfigBase = { apiVersion: this.$options.apiVersion, serverProto: this.$options.serverProto };
```
that way one would have autocomplete when trying to create the static object and in addition compilation would fail should you add any mandatory property to the `IServerConfigBase` interface in the future.

lib/commands/config/config-set.ts
```diff
@@ -1,12 +1,19 @@
...
+
+	constructor(private $nsCloudOptionsProvider: ICloudOptionsProvider,
+		private $nsCloudServerConfigManager: IServerConfigManager,
+		private $options: ICloudOptions,
+		private $stringParameterBuilder: IStringParameterBuilder) { }
+
 	public async execute(args: string[]): Promise<void> {
 		const configurationFile = args[0];
-		this.$nsCloudServerConfigManager.applyConfig(null, { configurationFile: configurationFile });
+		const globalServerConfig = { apiVersion: this.$options.apiVersion, serverProto: this.$options.serverProto };
```
Mitko-Kerezov: Should you decide to follow my suggestion from the previous comment it can be applied here as well

lib/commands/config/config-set.ts
```diff
@@ -1,12 +1,19 @@
...
+		this.$nsCloudServerConfigManager.applyConfig(null, { configurationFile: configurationFile }, globalServerConfig);
```
Mitko-Kerezov: This can be shortened to
```TypeScript
this.$nsCloudServerConfigManager.applyConfig(null, { configurationFile }, globalServerConfig);
```

lib/server-config-manager.ts
```diff
@@ -23,16 +23,16 @@ export class ServerConfigManager implements IServerConfigManager {
...
 		const newConfig = this.loadConfig(`${ServerConfigManager.CONFIG_FILE_NAME}-${configName}`, options);
-		this.mergeConfig(baseConfig, newConfig);
+		this.mergeConfig(baseConfig, newConfig, globalServerConfig);
 		this.saveConfig(baseConfig, ServerConfigManager.CONFIG_FILE_NAME);
 	}
 
 	public printConfigData(): void {
 		const config = this.getCurrentConfigData();
-		console.log(config);
+		console.log(JSON.stringify(config, null, "  "));
```
Mitko-Kerezov: Another personal preference - I'd prefer a number as the third argument of `JSON.stringify` - seems a bit more natural to me:
`console.log(JSON.stringify(config, null, 2));`

PR #88, opened 2018-01-29
Add project name in the cleanup output

TsvetanMilanov:
Add the project name in the cloud cleanup output. This was requested from
our QAs.


Mitko-Kerezov:
On a side note, do we need `cleanupTaskId` in the regular output? 
Wouldn't it be better as a `trace` message?

TsvetanMilanov:
If we hide the task id we are going to ask our users to send us logs with `--log trace` when they have problems with the cleanup. If we show the task id we are going to skip this step. We did the same thing in the cloud build output.

Mitko-Kerezov:
@TsvetanMilanov seems reasoable, I agree, let's keep the id for problem localization facilitation

PR #89, opened 2018-02-04
Fix getting runtime and cli versions for cloud builds

rosen-vladimirov:
Fix getting runtime and cli versions for cloud builds by reading the
package.json of the project just before sending the request. Currently
we are relying on the nativescriptData passed by the caller, but in case
this is the first build for this platform, the nativeScriptData contains
only the id of the project. However, before sending the build request,
we are adding the platform, so the package.json contains the required
version of the runtime. So instead of using some logic to determine the
runtime version, just read the package.json.
Also remove the default fallbacks for CLI and runtime (3.0.0). From now
on, whenever we are unable to find version that should be used in the
cloud, an error will be raised.
This fixes the case when the first cloud build is executed with version
3.0.x of CLI and runtime, when tns-core-modules version in package.json
is next.
This also fixes the case when first and second cloud build are executed
with different CLI and runtime - when you have CLI 3.4.x, but you are
trying to build a project (for a first time), that has tns-core-modules
3.3.x as dependency, the first cloud build will be executed with CLI
3.3.x and runtime 3.3.x, while the second one will be with 3.4.x

Add tests for versionService.

### DIFF
--- a/lib/services/version-service.ts
+++ b/lib/services/version-service.ts
@@ -1,9 +1,7 @@
-import { join } from "path";
 import * as semver from "semver";
 
 export class VersionService implements IVersionService {
-	constructor(private $fs: IFileSystem,
-		private $httpClient: Server.IHttpClient,
+	constructor(private $httpClient: Server.IHttpClient,
 		private $logger: ILogger,
 		private $projectDataService: IProjectDataService) { }
 
@@ -11,18 +9,12 @@ export class VersionService implements IVersionService {
 		try {
 			const latestMatchingVersion = process.env.TNS_CLI_CLOUD_VERSION || await this.getLatestMatchingVersion("nativescript", this.getVersionRangeWithTilde(runtimeVersion));
 			if (!latestMatchingVersion) {
-				throw new Error(`Cannot find matching CLI version based on runtime version ${runtimeVersion}`);
+				throw new Error("Cannot find CLI versions.");
 			}
 
 			return latestMatchingVersion;
 		} catch (err) {
-			this.$logger.trace(`Unable to get information about CLI versions. Error is: ${err.message}`);
-			if (!semver.valid(runtimeVersion)) {
-				// In case the runtime version is not a valid semver version, we cannot determine CLI version.
-				throw new Error("Unable to determine CLI version for cloud build.");
-			}
-
-			return `${semver.major(runtimeVersion)}.${semver.minor(runtimeVersion)}.0`;
+			throw new Error(`Unable to determine CLI version for cloud build based on project's runtime version: ${runtimeVersion}. Error is: ${err.message}`);
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lib/services/version-service.ts
```diff
@@ -2,41 +2,42 @@ import { join } from "path";
...
+
+			return latestMatchingVersion;
 		} catch (err) {
 			this.$logger.trace(`Unable to get information about CLI versions. Error is: ${err.message}`);
+			if (!semver.valid(runtimeVersion)) {
+				// In case the runtime version is not a valid semver version, we cannot determine CLI version.
+				throw new Error("Unable to determine CLI version for cloud build.");
+			}
+
 			return `${semver.major(runtimeVersion)}.${semver.minor(runtimeVersion)}.0`;
```
rosen-vladimirov: maybe we should remove this as well

test/services/version-service.ts
```diff
@@ -0,0 +1,194 @@
...
+			assert.isTrue(isGetNSValueCalled, "getNSValue must be called");
+		});
+
+		it("fails when runtime is not added to package.json", async () => {
+			const testInjector = createTestInjector();
+			const projectDataService = testInjector.resolve<IProjectDataService>("projectDataService");
+			const projectDir = "project dir";
+			const platform = "platform";
+			let isGetNSValueCalled = false;
+			projectDataService.getNSValue = (projDir: string, propertyName: string): any => {
```
rosen-vladimirov: these several lines are duplicated with the test above

lib/services/version-service.ts
Mitko-Kerezov: along with the whole `try ... catch` block. 
Let's just `throw` an error whenever there isn't a matching version and that's it

test/services/version-service.ts
```diff
@@ -0,0 +1,194 @@
...
+				const versionService = testInjector.resolve<IVersionService>(VersionService);
+				const actualCliVersion = await versionService.getCliVersion(runtimeVersion);
+				assert.equal(actualCliVersion, expectedCliVersion);
+				assert.isTrue(isHttpRequestCalled, "An http request to registry.npmjs.org should have been made in order to get data for CLI.");
+			});
+		});
+
+		it("fails when unable to determine CLI version as runtime version is not valid semver version", async () => {
+			const testInjector = createTestInjector();
+			const body = {
```
Mitko-Kerezov: Why bother to create an unused object - you can just call
`mockHttpRequest(testInjector, {});`
down below

lib/services/version-service.ts
```diff
@@ -2,41 +2,42 @@ import { join } from "path";
...
+
+		if (!runtimeVersion) {
+			throw new Error(`Unable to find runtime version for package ${runtimePackageName}.`);
 		}
 
-		return runtimeVersion || VersionService.DEFAULT_CLI_VERSION;
+		return runtimeVersion;
 	}
 
 	public getCoreModulesVersion(projectDir: string): string {
```
Mitko-Kerezov: Let's delete this method as it currently remains unused.
My vote goes towards removing redundant code and introducing it subsequently should the need arise

test/services/version-service.ts
rosen-vladimirov: Just to ensure there's some data.

PR #91, opened 2018-02-06
Check if build is running using exponential backoff

Mitko-Kerezov:
Instead of retrying two times with a second and a half interval retry using exponential backoff.
Use 8 retries each after double than the last's timeout. Start from a second and a half.

Ping @rosen-vladimirov


package.json
```diff
@@ -45,6 +45,7 @@
     "typescript": "2.4.1"
   },
   "dependencies": {
+    "@types/promise-retry": "1.1.1",
```
rosen-vladimirov: this should be devDependency

PR #92, opened 2018-02-06
Add option to use shared cloud

Mitko-Kerezov:
Add option to use shared cloud for all cloud operations.

Ping @rosen-vladimirov


lib/services/cloud-publish-service.ts
```diff
@@ -91,6 +91,9 @@ export class CloudPublishService extends CloudService implements ICloudPublishSe
 
 	private async publishCore(publishRequestData: IPublishRequestData, publishDataCore: IPublishDataCore, getError: (publishResult: IBuildServerResult, publishRequestData: IPublishRequestData) => any): Promise<void> {
 		publishRequestData.packagePaths = await this.getPreparePackagePaths(publishDataCore);
+		if (publishDataCore.sharedCloud) {
```
rosen-vladimirov: do we really need this if?

PR #93, opened 2018-02-06
Change cleanupProject arg to object

Mitko-Kerezov:
Instead of accepting a number of arguments accept a single one

### DIFF
--- a/lib/definitions/server/server-project-service.d.ts
+++ b/lib/definitions/server/server-project-service.d.ts
@@ -8,9 +8,20 @@ interface ICleanupProjectResponse {
 	warnings: Error[];
 }
 
-interface ICleanupProjectData {
+/**
+ * Describes data used for project cleaning passed from client.
+ */
+interface ICleanupProjectDataBase extends IProjectNameComposition {
+	/**
+	 * Application's identifier
+	 */
 	appIdentifier: string;
-	projectName: string;
+}
+
+/**
+ * Describes data used for project cleaning passed to server.
+ */
+interface ICleanupProjectData extends ICleanupProjectDataBase {
 	templateAppName: string;
 	projectCleanupId: string;
 }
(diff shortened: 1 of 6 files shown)


lib/definitions/server/server-project-service.d.ts
```diff
@@ -8,9 +8,20 @@ interface ICleanupProjectResponse {
...
+interface ICleanupProjectDataBase extends IProjectNameComposition {
+	/**
+	 * Application's identifier
+	 */
 	appIdentifier: string;
-	projectName: string;
+}
+
+/**
+ * Descirbes data used for project cleaning passed to server.
```
rosen-vladimirov: Describes

lib/services/cloud-project-service.ts
```diff
@@ -24,10 +24,10 @@ export class CloudProjectService extends CloudService implements ICloudProjectSe
 		return path.join(options.projectDir, "cleanup-results");
 	}
 
-	public async cleanupProject(appIdentifier: string, projectName: string): Promise<ICleanupProjectResult> {
+	public async cleanupProject(cleanupProjectData: ICleanupProjectDataBase): Promise<ICleanupProjectResult> {
 		const cleanupTaskId = uuid.v4();
 		try {
-			const result = await this.executeCleanupProject(cleanupTaskId, appIdentifier, projectName);
+			const result = await this.executeCleanupProject(cleanupTaskId, cleanupProjectData.appIdentifier, cleanupProjectData.projectName);
```
rosen-vladimirov: you can use destructuring on line 40:
```TypeScript
private async executeCleanupProject(cleanupTaskId: string, { appIdentifier: string, projectName: string }): Promise<ICleanupProjectResult>
```
and here call the method directly with cleanupProjectData:
```TypeScript
const result = await this.executeCleanupProject(cleanupTaskId, cleanupProjectData);
```

PR #101, opened 2018-03-07
fix(kinvey): Set correct url for Kinvey Free EULA

rosen-vladimirov:
Set the correct URL for Kinvey Free EULA. Currently it redirects to the Kinvey site, but it will be updated to contain a PDF file with the correct EULA


Fatme:
👍 Looks ok to me

PR #106, opened 2018-03-15
feat: Improve package.json, so CLI can detect this as extension

rosen-vladimirov:
In case extension is not installed, but user tries to use any command from it, CLI tries to find out in which extension the command is registered. It works by checking all packages in npm that have the `nativescript:extension` keyword.
After that it reads their package.json and searches for nativescript.commands array.
So prepare the `nativescript-cloud` by adding all commands to the correct place in package.json. Add required keyword and some addtional ones.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,5 +62,44 @@
     "temp": "0.8.3",
     "uuid": "3.0.1",
     "xml-escape": "1.1.0"
+  },
+  "keywords": [
+    "nativescript",
+    "nativescript:extension",
+    "nativescript-cloud",
+    "cloud",
+    "build"
+  ],
+  "nativescript": {
+    "commands": [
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -62,5 +62,44 @@
...
+      "account|features",
+      "account|usage",
+      "build|cloud",
+      "cloud|build",
+      "cloud|clean|workspace",
+      "cloud|codesign",
+      "cloud|deploy",
+      "cloud|lib|version",
+      "cloud|publish|android",
+      "cloud|publish|ios",
```
Fatme: maybe missing "publish|cloud" or it is not supported?

rosen-vladimirov: It is not supported

PR #116, opened 2018-05-18
Add policy service

TsvetanMilanov:
## V3:
`¯\_(:|)_/¯`

## V2:
Add `nsCloudPolicyService`. The service provides methods for:
- checking if policy with any name and content in local file is accepted.
- checking if the client should ask the user to accept a policy.
- accepting policy with any name and content in local file
- getting a policy local user setting.
- getting all policies local user settings.

Add `nsCloudServicesPolicyService`. The service provides methods for:
- getting the Progress Software Privacy Policy url.
- getting the Cloud Services Policy content.
- getting the submit a GDPR Data Subject Access Request url.
- accepting the Cloud Services Policy.
- checking if the Cloud Services Policy is accepted.

Add `tns policy accept` command.
Example usage:
```bash
tns policy accept privacy-policy
```
```bash
tns policy accept my-policy ~/policy.txt
```

Also we need to store which policies are accepted by the current user. The fastest solution is to just send the local user policies settings to our server on successful login. The endpoint for this is in the account service - `/consent`. It accepts the whole map with the settings.

### DIFF
--- a/lib/services/eula-service-base.ts
+++ b/lib/services/eula-service-base.ts
@@ -5,13 +5,14 @@ import * as temp from "temp";
 export abstract class EulaServiceBase implements IEulaService {
 	private isEulaDownloadedInCurrentProcess = false;
 
-	constructor(private $httpClient: Server.IHttpClient,
-		private $userSettingsService: IUserSettingsService,
-		private $logger: ILogger,
-		private $fs: IFileSystem,
-		private $nsCloudDateTimeService: IDateTimeService,
+	constructor(private $fs: IFileSystem,
+		private $httpClient: Server.IHttpClient,
 		private $lockfile: ILockFile,
-		private $settingsService: ISettingsService) { }
+		private $logger: ILogger,
+		private $nsCloudDateTimeService: IDateTimeService,
+		private $nsCloudHashService: IHashService,
+		private $settingsService: ISettingsService,
+		private $userSettingsService: IUserSettingsService) { }
 
 	// Exposed for Sidekick
 	public async getEulaData(): Promise<IEulaData> {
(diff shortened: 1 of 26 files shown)


lib/services/eula-service-base.ts
```diff
@@ -60,7 +61,7 @@ export abstract class EulaServiceBase implements IEulaService {
 
 		// At this point we should have already downloaded the EULA, so just get the info for local file.
 		// If it does not exist - we were unable to download it.
-		const currentEulaHash = await this.getLocalEulaHash();
+		const currentEulaHash = await this.$nsCloudHashService.getLocalFileHash(this.getPathToEula());
```
Mitko-Kerezov: Not a merge-stopper, but instead of this approach you could change the `getLocalEulaHash` implementation to 
```TypeScript
private async getLocalEulaHash(): Promise<string> {
	return this.$nsCloudHashService.getLocalFileHash(this.getPathToEula());
}
```

And leave both calls unaltered

lib/services/policy-service.ts
```diff
@@ -0,0 +1,70 @@
...
+		return currentContent !== acceptableContent;
+	}
+
+	public async shouldAskToAcceptPolicy(policy: string): Promise<boolean> {
+		const currentContent = await this.getPolicyUserSetting(policy);
+		return !currentContent;
+	}
+
+	public async accept(data: IAcceptPolicyData): Promise<void> {
+		let policyCloudContent = await this.getPolicyContentFromAcceptData(data);
```
Mitko-Kerezov: This could be `const`

lib/services/policy-service.ts
```diff
@@ -0,0 +1,70 @@
...
+		if (!policyCloudContent) {
+			this.$errors.failWithoutHelp("Invalid policy.");
+		}
+
+		await this.setPolicyUserSetting(data.policyName, policyCloudContent);
+	}
+
+	public async acceptInTheCloud(data: IAcceptPolicyData): Promise<void> {
```
Mitko-Kerezov: Unsure about what the purpose of this method is. The method `accept` above this has the exact same implementation

PR #117, opened 2018-05-28
Add If-Modified-Since header to EULA GET request

ventsislav-georgiev:
This would make the server return 301 Not Modified status without returning the file content in case we already have it downloaded.

### DIFF
--- a/test/services/eula-service.ts
+++ b/test/services/eula-service.ts
@@ -13,6 +13,9 @@ interface IEulaTestData {
 	isLocalEulaFileMissing?: boolean;
 	eulaLastModifiedTimeInEpoch?: number;
 	currentTimeInEpoch?: number;
+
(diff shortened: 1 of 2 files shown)


test/services/eula-service.ts
```diff
@@ -114,6 +122,33 @@ describe("eulaService", () => {
...
+		},
+		{
+			testName: "should send If-Modified-Since header with the local EULA mtime",
+			expectedHttpRequestIfModifiedSinceHeader: new Date(2018, 5).toUTCString(),
+			eulaLastModifiedTimeInEpoch: new Date(2018, 5).getTime(),
+		}
+	];
+
+	["getEulaData", "getEulaDataWithCache"].forEach(methodName => {
+		describe.only(methodName, () => {
```
rosen-vladimirov: forgotten `.only` here

PR #118, opened 2018-05-28
Generate and store password for the encrypted workspace.

TomaNikolov:
The password will be per user and per project. It will be stored in the new file: `encryption-settings.json`.


lib/services/cloud-build-service.ts
```diff
@@ -334,7 +335,8 @@ export class CloudBuildService extends CloudService implements ICloudBuildServic
 				flavorId: settings.projectSettings.flavorId,
 				additionalCliFlags: settings.additionalCliFlags,
 				useIncrementalBuild: !settings.projectSettings.clean,
-				userEmail: this.$nsCloudUserService.getUser().email
+				userEmail: this.$nsCloudUserService.getUser().email,
+				macOSImagePassword: await this.$nsCloudImageEncryptionService.getImagePassword(settings.projectSettings)
```
rosen-vladimirov: I prefer having this calculated before constructing the object as it will be easier for debugging, but this is just a personal preference, not a merge stopper.

rosen-vladimirov:
Just wondering if the passwords are so long, maybe we should introduce a new file instead of poluting the user-settings.json. 
You can do this by using the `path.join(settingsService.getProfileDir(), "cloud-configuration.json")`;
In fact we can join this PR with this one: https://github.com/NativeScript/nativescript-cloud/pull/110 and use the same file. But we can discuss this offline.

lib/services/factory-settings-service.ts
```diff
@@ -0,0 +1,18 @@
...
+export class FactorySettingsService implements IFactorySettingsService {
+	constructor(private $fs: IFileSystem,
+		private $settingsService: ISettingsService,
+		private $userSettingsService: IUserSettingsService,
+		private $lockfile: ILockFile,
+		private $logger: ILogger) {
+	}
+
+	public getUserSettingsService(): IUserSettingsService {
+		const userSettingServiceInstance = new (<any>this.$userSettingsService).__proto__.constructor(this.$fs, this.$settingsService, this.$lockfile, this.$logger);
```
TsvetanMilanov: `¯\_(:|)_/¯`

PR #126, opened 2018-07-03
Execute local prepare and cloud build in parallel.

TomaNikolov:
This way we will avoid waiting to prepare the project two times. Once locally and once in the cloud.


lib/services/cloud-build-service.ts
```diff
@@ -128,70 +201,83 @@ export class CloudBuildService extends CloudService implements ICloudBuildServic
...
+			packageJsonContent.nativescript = {};
+		}
+
+		if (!packageJsonContent.nativescript[runtimeName]) {
+			packageJsonContent.nativescript[runtimeName] = {
+				version: runtimeVersion
+			};
+
+			this.$fs.writeJson(packageJsonPath, packageJsonContent);
+		}
```
rosen-vladimirov: we need to handle the case where the `package.json` contains a tag instead of a version - for example `next` or `rc`. 
The current behavior in this case is:
1. The project preparation replaces the tag with a real version
2. Cloud build is started and the version in package.json is exact, not a tag.

With the changes in this PR, when the cloud build starts, we'll have a tag in the package.json and we may end up with different versions of runtime in the cloud and locally.

PR #127, opened 2018-07-04
Default errors string

Mitko-Kerezov:
In case server does not return an `errors` property we'll get an undefined later on when displaying the error. Mitigate that.

Ping @rosen-vladimirov


rosen-vladimirov:
Closing as we do not want the default.

PR #133, opened 2018-10-10
fix: Webpack variables are not transferred correctly during build

ggarabedian:
Related to https://github.com/NativeScript/sidekick-feedback/issues/277.


teobugslayer:
Ship it!

PR #134, opened 2018-10-16
fix: hmr is not working in some cases

rosen-vladimirov:
### fix: hmr is not working when git is used to upload files
Using cloud build with Hot Module Replacement is not working when git is used as it modifies the line endings. Most of the templates are published from Windows machine, so the line endings in them are `\r\n`.
The local prepare of the project produces bundle.js with `hotCurrentHash = <value 1>`. Once we use git to push the files in the cloud, the line endings are `\n`. The `bundle.js` produced in the cloud contains `hotCurrentHash = <value 2>`.
After the built .apk is downloaded from the cloud and installed on device, CLI does not apply full sync as it thinks the local files are the same as the ones in the cloud.
In order to fix this, set `core.autocrlf` to false in the local codecommit repo. This way the line endings will be the same and the `hotCurrentHash` should be the same in the cloud and locally. In order to force creation of new repo, add a versioning to the dir name of the local repo.
This allows us to force creation of new repo in this case.

### fix: do not generate nshashes file when --hmr is used
When cloud build is used with hmr, the files generated in the cloud (bundle.js especially) may differ from the local ones.
Currently we build the app in the cloud and generate hashes locally based on the local files, instead of getting them from the cloud. This works in normal case, but not for HMR.
Until we get the hashes from the cloud, skip generation of nshashes in case hmr is passed.

### DIFF
--- a/lib/services/git-service.ts
+++ b/lib/services/git-service.ts
@@ -11,6 +11,12 @@ export class GitService implements IGitService {
 	private static MAX_TIME_FOR_UNUSED_REPOSITORY = 1000 * 3600 * 24 * 30;
 	private static MINIMAL_GIT_MAJOR_VERSION = 2;
 	private static MINIMAL_GIT_MINOR_VERSION = 9;
+	/**
+	 * Used when we want to force creation of new local repository after release of nativescript-cloud package.
+	 * Just increment this version and new local repositories will be created for each project.
+	 * Required for cases where we want to change some git settings in the repo and ensure they are pushed in the cloud.
+	 */
+	private static GIT_REPO_VERSION = "v1";
 
 	private gitFilePath: string;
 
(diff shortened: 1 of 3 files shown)


lib/services/git-service.ts
```diff
@@ -150,6 +163,16 @@ export class GitService implements IGitService {
 		}
 	}
 
+	private async gitCheckIfAutocrlfIsFalse(projectSettings: INSCloudProjectSettings): Promise<boolean> {
+		try {
+			const result = await this.executeCommand(projectSettings, ["config", "core.autocrlf"]);
+			return result && result.stdout && result.stdout.toString().trim().toLowerCase() === "false";
+		} catch (err) {
+			this.$logger.trace("Error while checking if core.autcrlf is false: ", err);
```
rosen-vladimirov: this should be `core.autocrlf`

PR #138, opened 2018-10-31
feat: add validation for "--aab" build option

KristianDD:


lib/commands/cloud-build.ts
```diff
@@ -31,6 +32,7 @@ export class CloudBuildCommand extends BundleValidatorBaseCommand implements ICo
 	public async canExecute(args: string[]): Promise<boolean> {
 		this.$bundleValidatorHelper.validate();
 		await this.$nsCloudEulaCommandHelper.ensureEulaIsAccepted();
+		this.$androidBundleValidatorHelper.validateNoAab();
```
Fatme: Is this backwards compatible with old CLI versions?

PR #143, opened 2018-12-03
 Allow Kinvey Studio users to perform cloud builds

TomaNikolov:
* Add custom namespace in config
* Read and write kinvey user from/to kinvey studio user file.


lib/services/kinvey-user-service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export class KinveyUserService extends UserServiceBase implements IUserService {
+	constructor($injector: IInjector,
+		private $hostInfo: IHostInfo) {
+		super($injector);
+		this.userFilePath = this.getUserFilePath();
+	}
+
+	public getUserData(): IUserData {
+		const userData: IKinveyUserData = <any>super.getUserData();
```
Mitko-Kerezov: You can set the `getUserData` method's return type to `any` in order to avoid the cast

lib/services/kinvey-user-service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	public setUserData(userData: IUserData): void {
+		const kinveyUserData: any = {
+			email: userData.userInfo.email,
+			firstName: userData.userInfo.firstName,
+			lastName: userData.userInfo.lastName,
+			token: userData.accessToken
+		};
+
+		super.setUserData(kinveyUserData);
```
Mitko-Kerezov: this method's parameter can be of type `any` too

lib/services/user-service-base.ts
```diff
@@ -0,0 +1,84 @@
+export abstract class UserServiceBase {
+	private userData: IUserData;
+	protected userFilePath: string;
+
+	private get $nsCloudServerAccountsService(): IServerAccountsService {
+		return this.$injector.resolve<IServerAccountsService>("nsCloudServerAccountsService");
```
Mitko-Kerezov: IMO you can put all of the dependencies inside the class' `constructor` instead of having them all as properties.
Property dependencies exist primarily because of cyclic chains

lib/services/user-service.ts
```diff
@@ -1,89 +1,47 @@
...
-		const userInfo = await this.$nsCloudServerAccountsService.getUserInfo();
-		return userInfo.externalAvatarUrl;
+		return this.userService.getUserAvatar();
 	}
 
-	private getUserFilePath(): string {
-		return join(this.userInfoDirectory, ".nativescript-user");
+	private determineUserService(): IUserService {
+		const serverConfig = this.$nsCloudServerConfigManager.getCurrentConfigData();
+		const namespace = <string>serverConfig["namespace"];
```
Mitko-Kerezov: You can declare this as:
`const namespace: string = serverConfig["namespace"];`

instead of casting

PR #148, opened 2019-01-03
Use custom http-client temporary

TsvetanMilanov:
We need to remove the stuck request check because it works for GET
requests, but for POST/PUT, we won't receive response until the whole
payload is transfered. The request.on("request") event is emitted even
if the SSL handshake is not complete and that's why we will remove this
check for now.


lib/bootstrap.ts
```diff
@@ -1,5 +1,7 @@
 import * as path from "path";
 
+$injector.overrideAlreadyRequiredModule = true;
```
rosen-vladimirov: :crying_cat_face:

PR #156, opened 2019-01-25
Add nsCloudConfigManager to the public API

TsvetanMilanov:
Add `nsCloudConfigManager` to the public API. It should provide methods for managing the cloud configuration and for listing all cloud services domain names.

This PR also removes the unused cloud services.

### DIFF
--- a/lib/cloud-config-manager.ts
+++ b/lib/cloud-config-manager.ts
@@ -1,19 +1,19 @@
 import * as path from "path";
 
-export class ServerConfigManager implements ICloudConfigManager {
+export class CloudConfigManager implements ICloudConfigManager {
 	private static CONFIG_FILE_NAME: string = "config";
-	private static BASE_CONFIG_FILE_NAME: string = `${ServerConfigManager.CONFIG_FILE_NAME}-base`;
+	private static BASE_CONFIG_FILE_NAME: string = `${CloudConfigManager.CONFIG_FILE_NAME}-base`;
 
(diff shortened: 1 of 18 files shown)


lib/cloud-config-manager.ts
```diff
@@ -1,10 +1,10 @@
 import * as path from "path";
 
-export class ServerConfigManager implements IServerConfigManager {
+export class ServerConfigManager implements ICloudConfigManager {
```
tsvetie: Shouldn't you rename this to "CloudConfigManager"?

lib/cloud-config-manager.ts
```diff
@@ -39,6 +39,37 @@ export class ServerConfigManager implements IServerConfigManager {
...
+		if (_.has(serviceConfig, valueName)) {
+			return serviceConfig[valueName];
+		}
+
+		return <string>config[valueName];
+	}
+
+	public getCloudServicesDomainNames(): string[] {
+		const cloudServices = this.getCurrentConfigData().cloudServices;
+		return _(cloudServices)
```
tsvetie: Empty line above return?

TsvetanMilanov:
No need to test it. It works on my machine 😄

lib/cloud-config-manager.ts
TsvetanMilanov: Done.

lib/cloud-config-manager.ts
TsvetanMilanov: 🎣

PR #158, opened 2019-02-05
Websockets

TsvetanMilanov:
We need to provide a way to send 2fa codes from the users machine to our build machines to enable iOS publish with 2fa enabled.

To do this, we need to consume the new server API for WebSockets communication. This communication channel can create problems for some users and that's why we won't use it for all of our cloud operations. Currently, we will use it only for the dev-apple-login and the publish. All other cloud operations will be v1, for now.

When the build machine asks for the 2fa code during the publish process, the cli or the {N} Sidekick will prompt the user for it. The user input will be sent to the build machine through the WebSockets communication channel. Detailed description of the communication protocol can be found in the server wikis.

The new `tns cloud dev-apple-login` command returns base64 encoded apple session. This session can be used in CI/CD systems to bypass the 2fa code. It used to be valid for 30 days, but now it is valid for ~8 hours. This duration is controlled by Apple and we can't change it.

The `tns cloud publish ios` command now accepts `--appleApplicationSpecificPassword` and `--appleSessionBase64` flags. The `--appleApplicationSpecificPassword` is another thing which is required by Apple when 2fa is enabled. If it is not provided, the cli or the {N} Sidekick will prompt the user for it. The `--appleSessionBase64` flag is used to provide the base64 encoded apple session to bypass the 2fa code prompt.


lib/cloud-operation/communication/communication-channel-base.ts
```diff
@@ -0,0 +1,120 @@
...
+
+		this.isConnected = false;
+		this.emit(CloudCommunicationEvents.CLOSE, code, reason);
+	}
+
+	protected abstract connectCore(): Promise<void>;
+	protected abstract closeCore(code: number, reason?: string): Promise<void>;
+	protected abstract sendMessageCore<T>(message: ICloudOperationMessage<T>): Promise<string>;
+
+	protected _handleMessage(m: any): ICloudOperationMessage<any> {
```
tsvetie: Remove the "_" prefix

lib/services/cloud-service.ts
```diff
@@ -1,89 +1,107 @@
...
-	protected async getContentOfS3File(pathToFile: string): Promise<string> {
-		return (await this.$httpClient.httpRequest(pathToFile)).body;
+	protected abstract getServerResults(serverResult: ICloudOperationResult): IServerItem[];
+
+	protected async executeCloudOperation<T>(cloudOperationName: string, action: (cloudOperationId: string) => Promise<T>): Promise<T> {
+		const cloudOperationId: string = v4();
+		try {
+			this.$logger.info(`Starting ${cloudOperationName}. Cloud operation id: ${cloudOperationId}`);
+			const result = await action(cloudOperationId);
+			return result
```
tsvetie: semicolon

lib/commands/interactive-cloud-command.ts
```diff
@@ -0,0 +1,44 @@
...
+export abstract class InteractiveCloudCommand implements ICommand {
+	public allowedParameters: ICommandParameter[];
+
+	constructor(private interactiveService: ICloudOperationService,
+		protected $errors: IErrors,
+		protected $logger: ILogger,
+		protected $prompter: IPrompter) { }
+
+	public async execute(args: string[]): Promise<void> {
+		this.interactiveService.on(CloudCommunicationEvents.MESSAGE, async (msg: ICloudOperationMessage<any>) => {
```
tsvetie: Not sure if this is applicable, as commands in CLI are not long-running and I guess the process will exit once the command completes. But I think we do not remove this handler anywhere

TsvetanMilanov: It won't hurt to remove it on precess exit.

lib/cloud-operation/cloud-operation-base.ts
```diff
@@ -0,0 +1,54 @@
...
+			throw err;
+		}
+	}
+
+	public async waitForResult(): Promise<ICloudOperationResult> {
+		this.isInitialized();
+		try {
+			const result = await this.waitForResultCore();
+
+			return result;
```
tsvetie: Shouldn't you set this.result here?

lib/cloud-operation/cloud-operation-factory.ts
```diff
@@ -0,0 +1,19 @@
+export class CloudOperationFactory implements ICloudOperationFactory {
```
tsvetie: [Suggestion] I don't really see the point in ICloudOperationFactory, ICloudCommunicationChannelFactory and IWebSocketFactory - I think you can use IFactory for all three

lib/cloud-operation/communication/communication-channel-base.ts
```diff
@@ -0,0 +1,120 @@
...
+				cloudOperationId: this.cloudOperationId,
+				body: {
+					code,
+					reason
+				}
+			};
+			await this.sendMessage<ICloudOperationStop>(stopMsg);
+			await this.closeCore(code, reason);
+		} catch (err) {
+			this.$logger.trace(err);
```
tsvetie: Why do we need the try .. catch here?

lib/cloud-operation/communication/communication-channel-base.ts
```diff
@@ -0,0 +1,120 @@
...
+export abstract class CommunicationChannelBase extends EventEmitter implements ICloudCommunicationChannel {
+	private static readonly PING_INTERVAL: number = 10000;
+	private static readonly MISSING_PING_REPLIES_COUNT: number = 6;
+
+	protected isConnected: boolean;
+	private pingInterval: NodeJS.Timer;
+	private echoSequenceNumber: number = 0;
+	private echoReplySequenceNumber: number = 0;
+
+	private handshakeCompleteResolve: (value?: PromiseLike<void>) => void;
```
tsvetie: [Suggestion] handshakeResolve

lib/cloud-operation/communication/websocket-channel.ts
```diff
@@ -0,0 +1,76 @@
...
+			});
+
+			this.addChannelListeners();
+		});
+	}
+
+	private addChannelListeners() {
+		this.ws.on(CloudCommunicationEvents.MESSAGE, m => {
+			const msg = super.handleMessage(m);
+			this.emit(CloudCommunicationEvents.MESSAGE, msg);
```
tsvetie: [Suggestion] Move this.emit to handleMessage

lib/cloud-operation/communication/communication-channel-base.ts
```diff
@@ -0,0 +1,120 @@
...
+		} else {
+			msg = m;
+		}
+
+		if (msg.type === CloudOperationMessageTypes.CLOUD_OPERATION_SERVER_HELLO) {
+			this.handshakeCompleteResolve();
+		} else if (msg.type === CloudOperationMessageTypes.CLOUD_OPERATION_ECHO_REPLY) {
+			const body: ICloudOperationEcho = msg.body;
+			if (body.identifier === this.cloudOperationId) {
+				this.echoReplySequenceNumber = body.sequenceNumber;
```
tsvetie: Does it matter which exactly is the sequenceNumber? That is, can we just reset the counter here?

lib/cloud-operation/communication/websocket-channel.ts
```diff
@@ -0,0 +1,76 @@
...
+					resolve(id);
+				});
+			} catch (err) {
+				reject(err);
+			}
+		});
+	}
+
+	protected async connectCore(): Promise<void> {
+		this.ws = this.$nsCloudWebSocketFactory.create(this.data.config.url);
```
tsvetie: Where do you set this.data?

lib/cloud-operation/communication/websocket-channel.ts
```diff
@@ -0,0 +1,76 @@
...
+
+			this.ws.once("close", closeHandler);
+			this.ws.once("unexpected-response", unexpectedResponseHandler);
+			this.ws.once("open", () => {
+				this.ws.removeListener("close", closeHandler);
+				this.ws.removeListener("unexpected-response", unexpectedResponseHandler);
+				resolve();
+			});
+
+			this.addChannelListeners();
```
tsvetie: Can we move addChannelListeners to the "open" handler or is there a change to miss some messages?

lib/cloud-operation/communication/communication-channel-base.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	constructor(protected cloudOperationId: string,
+		protected data: ICloudCommunicationChannelData<any>,
+		protected $logger: ILogger) {
+		super();
+	}
+
+	public async connect(): Promise<void> {
+		if (!this.isConnected) {
+			await this.connectCore();
```
tsvetie: connectCore can be rejected - do we need try .. catch?

lib/cloud-operation/communication/websocket-channel.ts
```diff
@@ -0,0 +1,76 @@
...
+		this.ws = this.$nsCloudWebSocketFactory.create(this.data.config.url);
+		await new Promise<void>((resolve, reject) => {
+			const closeHandler = (c: number, r: string) => {
+				reject(new Error(`Connection closed with code: ${c}`));
+				this.close(c, r);
+			};
+			const unexpectedResponseHandler = () => {
+				const errMsg = "Unexpected response received.";
+				reject(new Error(errMsg));
+				this.close(CloudCommunicationChannelExitCodes.UNEXPECTED_RESPONSE, errMsg);
```
tsvetie: [Suggesction] Move this.close calls to super.connect

lib/cloud-operation/communication/websocket-channel.ts
```diff
@@ -0,0 +1,76 @@
...
+		});
+	}
+
+	private addChannelListeners() {
+		this.ws.on(CloudCommunicationEvents.MESSAGE, m => {
+			const msg = super.handleMessage(m);
+			this.emit(CloudCommunicationEvents.MESSAGE, msg);
+		});
+
+		this.ws.once("error", err => this.emit("error", err));
```
tsvetie: Should we call this.close on error? If not, then why do we use once?

lib/cloud-operation/communication/websocket-channel.ts
```diff
@@ -0,0 +1,76 @@
...
+	constructor(protected cloudOperationId: string,
+		protected data: ICloudCommunicationChannelData<IWebSocketCloudChannelConfigProperties>,
+		protected $logger: ILogger,
+		private $nsCloudWebSocketFactory: IWebSocketFactory) {
+		super(cloudOperationId, data, $logger);
+	}
+
+	protected async closeCore(code: number, reason?: string): Promise<void> {
+		if (this.ws.readyState === this.ws.OPEN) {
+			this.ws.removeAllListeners();
```
tsvetie: I think you should move this outside of the IF - e.g. if connectCore fails, who will remove the open handler.

lib/cloud-options-provider.ts
```diff
@@ -6,13 +6,15 @@ export class CloudOptionsProvider implements ICloudOptionsProvider {
...
 			local: { type: OptionType.Boolean, hasSensitiveValue: false },
-			serverProto: { type: OptionType.String , hasSensitiveValue: true},
+			serverProto: { type: OptionType.String, hasSensitiveValue: true },
 			namespace: { type: OptionType.String, hasSensitiveValue: true },
 			instanceId: {type: OptionType.String, hasSensitiveValue: false },
 			sharedCloud: { type: OptionType.Boolean, hasSensitiveValue: false },
 			workflow: { type: OptionType.Object, hasSensitiveValue: true },
 			vmTemplateName: { type: OptionType.String, hasSensitiveValue: false },
 			track: { type: OptionType.String, default: DEFAULT_ANDROID_PUBLISH_TRACK, hasSensitiveValue: false },
+			appleApplicationSpecificPassword: { type: OptionType.String, hasSensitiveValue: true },
```
tsvetie: I'm not sure where to check if this is used.

lib/cloud-operation/communication/communication-channel-base.ts
```diff
@@ -0,0 +1,120 @@
...
+		if (_.isBuffer(m)) {
+			msg = JSON.parse(m.toString("utf8"));
+		} else if (_.isString(m)) {
+			msg = JSON.parse(m);
+		} else {
+			msg = m;
+		}
+
+		if (msg.type === CloudOperationMessageTypes.CLOUD_OPERATION_SERVER_HELLO) {
+			this.handshakeCompleteResolve();
```
tsvetie: Should we have a reasonable timeout, after which if we have not received the serverHello, connect should fail?

lib/commands/interactive-cloud-command.ts
```diff
@@ -0,0 +1,53 @@
...
+					await this.interactiveService.sendCloudMessage(cloudMessage);
+				} catch (err) {
+					this.$logger.trace(`Can't send input to cloud operation ${msg.cloudOperationId}. Error is ${err}.`);
+				}
+			}
+		};
+
+		this.interactiveService.on(CloudCommunicationEvents.MESSAGE, messageHandler);
+
+		this.$processService.attachToProcessExitSignals(this, () => {
```
tsvetie: Does this mean we will remove the handlers when the whole process exits? If so, then if hypothetically I call the same command twice, I will end up with two handlers attached

server-configs/config.json
```diff
@@ -1,29 +1,29 @@
 {
-	"apiVersion": "v3",
+	"apiVersion": "pr174",
```
tsvetie: Should this be in the PR? (:

lib/services/cloud-service.ts
```diff
@@ -1,89 +1,108 @@
...
-	// TODO: Remove after 10.05.2018
-	private static OPERATION_IN_PROGRESS_STATUS_OLD = "Building";
-	private static OPERATION_IN_PROGRESS_STATUS = "InProgress";
-
-	protected outputCursorPosition: number;
+import { v4 } from "uuid";
+
+import { CloudOperationMessageTypes, CloudCommunicationEvents } from "../constants";
+
+export abstract class CloudService extends EventEmitter implements ICloudOperationService {
```
tsvetie: [Suggestion] ICloudOperationService => ICloudService

lib/commands/build-command-helper.ts
```diff
@@ -90,7 +90,19 @@ export class BuildCommandHelper implements IBuildCommandHelper {
...
 
-		return { username, password };
+		return {
+			username,
+			password
+		};
+	}
+
+	public async getExtendedAppleCredentials(args: string[], options: ICloudOptions): Promise<IPublishCredentials> {
+		const credentials = await this.getAppleCredentials(args);
```
tsvetie: We should not ask for getAppleCredentials if options.appleSessionBase64

lib/services/cloud-service.ts
```diff
@@ -97,4 +116,30 @@ export abstract class CloudService extends EventEmitter {
...
+
+		return null;
+	}
+
+	private async cleanup(): Promise<void> {
+		await Promise.all(_(this.cloudOperations)
+			.values<ICloudOperation>()
+			.map(op => op.cleanup())
+			.value()
+		);
```
tsvetie: Empty this.cloudOperations

lib/services/cloud-service.ts
```diff
@@ -97,4 +116,30 @@ export abstract class CloudService extends EventEmitter {
...
+		} catch (err) {
+			this.$logger.warn(`Unable to get cloud operation output. Error is: ${err}`);
+		}
+	}
+
+	protected getResult(cloudOperationId: string): ICloudOperationResult {
+		try {
+			return this.cloudOperations[cloudOperationId].getResult();
+		} catch (err) {
+			this.$logger.trace(err);
```
tsvetie: Shouldn't we throw the error?

lib/services/cloud-service.ts
```diff
@@ -97,4 +116,30 @@ export abstract class CloudService extends EventEmitter {
 
 		return targetFileNames;
 	}
+
+	protected async getCollectedLogs(serverResponse: IServerResponse): Promise<string> {
+		try {
+			return this.$nsCloudOutputFilter.filter(await this.$nsCloudS3Service.getContentOfS3File(serverResponse.outputUrl));
+		} catch (err) {
+			this.$logger.warn(`Unable to get cloud operation output. Error is: ${err}`);
```
tsvetie: Shouldn't we return ""?

lib/services/cloud-service.ts
```diff
@@ -1,89 +1,108 @@
...
-					resolve(await this.getObjectFromS3File<IServerStatus>(serverInformation.statusUrl));
-				} catch (err) {
-					this.$logger.trace(err);
-					reject(new Error(this.failedToStartError));
+	protected async waitForServerOperationToFinish(cloudOperationId: string, serverResponse: IServerResponse, options: ICloudOperationExecutionOptions): Promise<ICloudOperationResult> {
+		const cloudOperationVersion = serverResponse.cloudOperationVersion || CloudService.CLOUD_OPERATION_VERSION_1;
+		const cloudOperation: ICloudOperation = this.$nsCloudOperationFactory.create(cloudOperationVersion, cloudOperationId, serverResponse);
+		this.cloudOperations[cloudOperationId] = cloudOperation;
+
+		cloudOperation.on(CloudCommunicationEvents.MESSAGE, (m: ICloudOperationMessage<any>) => {
```
tsvetie: This handler is never removed

lib/services/cloud-service.ts
```diff
@@ -1,89 +1,108 @@
...
+		private $nsCloudOutputFilter: ICloudOutputFilter,
+		private $processService: IProcessService) {
 		super();
+		this.cloudOperations = Object.create(null);
+		this.$processService.attachToProcessExitSignals(this, this.cleanup.bind(this));
 	}
 
-	protected async getObjectFromS3File<T>(pathToFile: string): Promise<T> {
-		return JSON.parse(await this.getContentOfS3File(pathToFile));
+	public abstract getServerOperationOutputDirectory(options: IOutputDirectoryOptions): string;
```
tsvetie: Can we make getServerOperationOutputDirectory and getServerResults virtual - returning "" and [] respectively. If I am correct, these are not required by all services

lib/services/cloud-service.ts
```diff
@@ -1,89 +1,108 @@
...
+			await cloudOperation.init();
+		} catch (err) {
+			this.$logger.trace(err);
+			await cloudOperation.cleanup();
+			throw new Error(this.failedToStartError);
+		}
+
+		try {
+			const result = await cloudOperation.waitForResult();
+			await cloudOperation.cleanup();
```
tsvetie: When the operation completes, should we remove it from this.cloudOperations?

lib/services/cloud-build-service.ts
```diff
@@ -145,13 +146,12 @@ export class CloudBuildService extends CloudService implements ICloudBuildServic
...
 
-		this.emitStepChanged(buildId, constants.BUILD_STEP_NAME.BUILD, constants.BUILD_STEP_PROGRESS.START);
+		this.emitStepChanged(cloudOperationId, constants.BUILD_STEP_NAME.BUILD, constants.BUILD_STEP_PROGRESS.START);
 		const buildResponse: IServerResponse = await this.$nsCloudServerBuildService.startBuild(buildProps);
 		this.$logger.trace("Build response:");
 		this.$logger.trace(buildResponse);
-		await this.waitForServerOperationToFinish(buildId, buildResponse);
-		const buildResult: IBuildServerResult = await this.getObjectFromS3File<IBuildServerResult>(buildResponse.resultUrl);
-		this.emitStepChanged(buildId, constants.BUILD_STEP_NAME.BUILD, constants.BUILD_STEP_PROGRESS.END);
+		const buildResult: ICloudOperationResult = await this.waitForServerOperationToFinish(cloudOperationId, buildResponse, { silent: false });
```
tsvetie: In the previous code, we used to download the results (getObjectFromS3File), but now, we are not

lib/services/cloud-project-service.ts
```diff
@@ -87,26 +86,40 @@ export class CloudProjectService extends CloudService implements ICloudProjectSe
...
+					return { output, err: `${childCloudOperationId} error: ${err}` };
 				}
+			};
+
+			childTasks.push(executeChildTask());
+		}
+
+		for (let task of childTasks) {
+			const taskResult = await task;
+			this.$logger.info(taskResult.output);
```
tsvetie: Why not print the output and the errors in the previous FOR cycle?

lib/services/cloud-codesign-service.ts
```diff
@@ -111,30 +113,27 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
...
 			stdout: codesignResult.stdout,
-			fullOutput: fullOutput,
 			outputFilesPaths: localCodesignResults
 		};
 
 		return result;
 	}
 
-	protected async getServerLogs(logsUrl: string, buildId: string): Promise<void> {
+	protected async getServerLogs(logsUrl: string, cloudOperationId: string): Promise<void> {
```
tsvetie: delete this

lib/services/cloud-publish-service.ts
```diff
@@ -18,112 +16,128 @@ export class CloudPublishService extends CloudService implements ICloudPublishSe
...
 		this.$logger.info("Publishing finished successfully.");
 	}
 
-	protected getServerResults(codesignResult: IBuildServerResult): IServerItem[] {
+	protected getServerResults(codesignResult: ICloudOperationResult): IServerItem[] {
 		return [];
 	}
 
-	protected async getServerLogs(logsUrl: string, buildId: string): Promise<void> {
+	protected async getServerLogs(logsUrl: string, cloudOperationId: string): Promise<void> {
```
tsvetie: delete this

lib/services/cloud-codesign-service.ts
```diff
@@ -69,22 +67,26 @@ export class CloudCodesignService extends CloudService implements ICloudCodesign
...
+		const codesignRequest = await this.prepareCodesignRequest(cloudOperationId, codesignData, projectData);
 		const codesignResponse: IServerResponse = await this.$nsCloudServerBuildService.generateCodesignFiles(codesignRequest);
 		this.$logger.trace(`Codesign response: ${JSON.stringify(codesignResponse)}`);
-
+		let codesignResult;
 		try {
-			await this.waitForServerOperationToFinish(buildId, codesignResponse);
+			codesignResult = await this.waitForServerOperationToFinish(cloudOperationId, codesignResponse, { silent: false });
 		} catch (ex) {
+			codesignResult = this.getResult(cloudOperationId);
```
tsvetie: [Suggestion] Move this.getResult to the catch in waitForServerOperationToFinish

lib/services/cloud-service.ts
```diff
@@ -97,4 +116,30 @@ export abstract class CloudService extends EventEmitter {
 
 		return targetFileNames;
 	}
+
+	protected async getCollectedLogs(serverResponse: IServerResponse): Promise<string> {
```
tsvetie: I think you should move this to CloudOperationV1 - doesn't it apply only to v1 operations?

lib/services/cloud-project-service.ts
```diff
@@ -87,26 +86,40 @@ export class CloudProjectService extends CloudService implements ICloudProjectSe
...
+				const childCloudOperationId = task.cloudOperationId;
+				let output = `Child cloud operation id: ${childCloudOperationId}`;
+				try {
+					const taskResult = await this.waitForServerOperationToFinish(childCloudOperationId, task, { silent: true });
+					tasksResults[childCloudOperationId] = taskResult;
+					output += await this.getCollectedLogs(task);
+					return { output };
+				} catch (err) {
+					if (this.getResult(childCloudOperationId)) {
+						tasksResults[childCloudOperationId] = this.getResult(childCloudOperationId);
```
tsvetie: Extract this.getResult(childCloudOperationId) to a const in order to remove repetition

lib/services/cloud-project-service.ts
```diff
@@ -87,26 +86,40 @@ export class CloudProjectService extends CloudService implements ICloudProjectSe
...
-			const task = cleanupResponse.buildMachineResponse[i];
-			try {
-				await this.waitForServerOperationToFinish(taskId, task);
-				const taskResult = await this.getObjectFromS3File<IServerResult>(task.resultUrl);
-				const output = this.$nsCloudOutputFilter.filter(await this.getContentOfS3File(task.outputUrl));
-				if (this.hasContent(output)) {
-					this.$logger.info(output);
+		const childTasks: Promise<IChildTaskExecutionResult>[] = [];
+		for (let task of cleanupResponse.buildMachineResponse) {
+			const executeChildTask = async (): Promise<IChildTaskExecutionResult> => {
```
tsvetie: No need to define the function every time - you can extract it as a service method, e.g. waitForChildTaskToFinish

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -0,0 +1,127 @@
...
+		private $logger: ILogger,
+		private $nsCloudOutputFilter: ICloudOutputFilter,
+		private $nsCloudS3Service: IS3Service) {
+		super(id, serverResponse);
+
+		this.outputCursorPosition = 0;
+	}
+
+	public async sendMessage<T>(message: ICloudOperationMessage<T>): Promise<void> {
+		this.$logger.warn("This version of the cloud operation does not support sending messages to the cloud services.");
```
tsvetie: Shouldn't this be an error?

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -0,0 +1,127 @@
...
+			}
+
+			if (this.serverStatus.status === CloudOperationBase.OPERATION_IN_PROGRESS_STATUS) {
+				await this.getServerLogs(this.serverResponse.outputUrl);
+			}
+
+		}, CloudOperationV1.OPERATION_STATUS_CHECK_INTERVAL);
+	}
+
+	private async getServerLogs(logsUrl: string): Promise<void> {
```
tsvetie: Can logsUrl be something other than this.serverResponse.outputUrl?

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -0,0 +1,127 @@
...
+						resolve(this.serverStatus);
+					})
+					.catch(err => {
+						this.$logger.trace(err);
+						reject(new Error("Failed to start."));
+					});
+			}).catch(retry);
+		}, promiseRetryOptions);
+
+		this.pollForLogs();
```
tsvetie: {Suggestion] Move this to the .then

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -0,0 +1,127 @@
...
+		this.logsCheckInterval = setInterval(async () => {
+			if (this.serverStatus.status === CloudOperationBase.OPERATION_COMPLETE_STATUS) {
+				await this.getServerLogs(this.serverResponse.outputUrl);
+				clearInterval(this.logsCheckInterval);
+				return;
+			}
+
+			if (this.serverStatus.status === CloudOperationBase.OPERATION_FAILED_STATUS) {
+				await this.getServerLogs(this.serverResponse.outputUrl);
+				return;
```
tsvetie: Shouldn't we clearInterval(this.logsCheckInterval);?

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -0,0 +1,127 @@
...
+
+				if (this.serverStatus.status === CloudOperationV1.OPERATION_FAILED_STATUS) {
+					try {
+						this.result = await this.$nsCloudS3Service.getJsonObjectFromS3File<ICloudOperationResult>(this.serverResponse.resultUrl);
+					} catch (err) {
+						this.$logger.trace(err);
+					}
+
+					clearInterval(this.statusCheckInterval);
+					return reject(new Error("Cloud operation failed"));
```
tsvetie: I think we should reject only if !this.result

lib/cloud-operation/cloud-operation-v2.ts
```diff
@@ -0,0 +1,94 @@
...
+			resolve();
+		});
+	}
+
+	protected waitForResultCore(): Promise<ICloudOperationResult> {
+		return this.waitResultPromise;
+	}
+
+	private subscribeForMessages(): Promise<ICloudOperationResult> {
+		let isResolved = false;
```
tsvetie: You can remove isResolved and use this.status instead - reject only if status is not OPERATION_COMPLETE_STATUS or OPERATION_FAILED_STATUS

lib/cloud-operation/cloud-operation-v2.ts
```diff
@@ -0,0 +1,94 @@
...
+		await this.communicationChannel.sendMessage(message);
+	}
+
+	public async cleanup(exitCode?: number): Promise<void> {
+		if (this.waitToStartTimeout) {
+			clearTimeout(this.waitToStartTimeout);
+			this.waitToStartTimeout = null;
+		}
+
+		await this.communicationChannel.close(exitCode);
```
tsvetie: Shouldn't we remove the attached handlers as well?

lib/cloud-operation/cloud-operation-v2.ts
```diff
@@ -0,0 +1,94 @@
...
+			this.communicationChannel.removeListener(CloudCommunicationEvents.CLOSE, closeHandler);
+			resolve();
+		});
+	}
+
+	protected waitForResultCore(): Promise<ICloudOperationResult> {
+		return this.waitResultPromise;
+	}
+
+	private subscribeForMessages(): Promise<ICloudOperationResult> {
```
tsvetie: [Suggestion] Rename to getResultPromise

lib/cloud-operation/cloud-operation-v2.ts
```diff
@@ -0,0 +1,94 @@
...
+			};
+			try {
+				this.communicationChannel.once(CloudCommunicationEvents.CLOSE, closeHandler);
+				await this.communicationChannel.connect();
+				this.status = CloudOperationV2.OPERATION_IN_PROGRESS_STATUS;
+			} catch (err) {
+				return reject(err);
+			}
+
+			this.waitResultPromise = this.subscribeForMessages();
```
tsvetie: I think we should do this after connect completes successfully

lib/cloud-operation/cloud-operation-v2.ts
```diff
@@ -0,0 +1,94 @@
...
+			}, CloudOperationV2.WAIT_TO_START_TIMEOUT);
+
+			const closeHandler = (code: number) => {
+				this.cleanup(code);
+				reject(new Error(`Communication channel closed with code ${code}`));
+			};
+			try {
+				this.communicationChannel.once(CloudCommunicationEvents.CLOSE, closeHandler);
+				await this.communicationChannel.connect();
+				this.status = CloudOperationV2.OPERATION_IN_PROGRESS_STATUS;
```
tsvetie: [Could] Not a big deal, but can we have this.status and this.waitResultPromise setters on subsequent lines (:

lib/cloud-operation/cloud-operation-v2.ts
```diff
@@ -0,0 +1,94 @@
...
+			} catch (err) {
+				return reject(err);
```
tsvetie: If cleanup does not remove listeners, then remove the CLOSE handler here as well

lib/services/cloud-build-service.ts
tsvetie: Ignore this: I was wrong

lib/services/cloud-service.ts
tsvetie: Edit: Move to cloud-operation-base

lib/services/s3-service.ts
```diff
@@ -0,0 +1,13 @@
+export class S3Service implements IS3Service {
```
tsvetie: [Suggestion] Rename to a helper

tsvetie:
Final check which operations should be silent

server-configs/config.json
TsvetanMilanov: No, but the apiVersion should be bumped to v4.

lib/commands/build-command-helper.ts
TsvetanMilanov: 🎣

lib/services/cloud-service.ts
TsvetanMilanov: We use this method only if there is error. If there is no result (the result file is empty or we can't get it), we want to throw the original error, not the getResult error. If we throw the error here, we need to trace the getResult error in every cloud service which uses this method. Do you think it will be better to throw the error here?

lib/services/cloud-service.ts
TsvetanMilanov: 1. I will remove all handlers in the cloud operation cleanup method.
2. We should always attach the handler because we are going to receive messages other than output and we need to proxy them. For example the inputRequest message.

lib/services/cloud-service.ts
TsvetanMilanov: Yes we should.

lib/services/cloud-service.ts
TsvetanMilanov: Done.

lib/services/cloud-service.ts
TsvetanMilanov: Done.

lib/services/cloud-service.ts
TsvetanMilanov: Done.

lib/services/cloud-project-service.ts
TsvetanMilanov: Because we need to start all cleanup tasks and then wait for their results and print their logs. If we wait the result of each cleanup task in the previous FOR cycle, the handshake between the build machine and the client application may fail for some tasks for cloud operation v2. Also if we print the logs there, we will receive mixed output from all tasks for cloud operation v2.

lib/services/cloud-codesign-service.ts
TsvetanMilanov: Done.

lib/services/cloud-publish-service.ts
TsvetanMilanov: Done.

lib/services/cloud-codesign-service.ts
TsvetanMilanov: Not all cloud services need the result in case of error and if we move it there, we are going to loose the error.

lib/services/cloud-service.ts
TsvetanMilanov: Done.

lib/services/cloud-project-service.ts
TsvetanMilanov: Done.

lib/cloud-operation/cloud-operation-v1.ts
TsvetanMilanov: No.

lib/cloud-operation/cloud-operation-v1.ts
TsvetanMilanov: It should.

lib/cloud-operation/cloud-operation-v1.ts
TsvetanMilanov: Done.

lib/cloud-operation/cloud-operation-v1.ts
TsvetanMilanov: Yes, and if we do this, we can remove all try/catch logic from the cloud services.

 🎣

lib/cloud-operation/cloud-operation-v2.ts
TsvetanMilanov: Done.

lib/cloud-operation/cloud-operation-v2.ts
TsvetanMilanov: 1. Yes, we should.
2. I removed the invocation of the cleanup method from within the cloud operation. Now it should be called by the user of the cloud operation instance. Do you think this is ok?

lib/cloud-operation/cloud-operation-v2.ts
TsvetanMilanov: Done.

lib/cloud-operation/cloud-operation-v2.ts
TsvetanMilanov: especially for you

lib/cloud-operation/cloud-operation-v2.ts
TsvetanMilanov: It does now.

lib/services/s3-service.ts
TsvetanMilanov: 1. Done
2. I prefer not to move the httpRequest call from downloadServerResults in the s3 helper because we can have results not related to s3.

TsvetanMilanov:
Only the cloud build service shouldn't be silent.

lib/cloud-operation/communication/communication-channel-base.ts
TsvetanMilanov: Because we may be in a situation where we can't send stop message and we don't want to stop any other cleanup tasks.

lib/cloud-operation/communication/communication-channel-base.ts
TsvetanMilanov: Done.

lib/cloud-operation/communication/websocket-channel.ts
TsvetanMilanov: Done.

lib/cloud-operation/communication/websocket-channel.ts
TsvetanMilanov: In the communication-channel-factory.

lib/cloud-operation/communication/websocket-channel.ts
TsvetanMilanov: It should be in the "open" handler 🎣

lib/cloud-operation/communication/communication-channel-base.ts
TsvetanMilanov: No. We want to throw the error.

lib/cloud-operation/communication/websocket-channel.ts
TsvetanMilanov: If we do this, we are going to loose the code and reason and we are only going to have err.

lib/cloud-operation/communication/websocket-channel.ts
TsvetanMilanov: We should use on.

lib/cloud-operation/communication/websocket-channel.ts
TsvetanMilanov: You are correct.

lib/cloud-operation/communication/communication-channel-base.ts
TsvetanMilanov: Yes, it should be 60 seconds. 30 seconds for the max AWS lambda duration and 30 seconds for the websocket connection.

lib/commands/interactive-cloud-command.ts
TsvetanMilanov: Yes and no. The processService attaches only one listener to the process exit signals and it will execute the provided functions in it. So technically, we will have only one listener and we won't receive the memory leak message, but we will add more than one function which will be executed on process exit. This is not a problem because you need to call execute more than once to add more listeners and we want to remove the listener attached in the specific execute call, not all listeners.

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	protected async initCore(): Promise<void> {
+		const promiseRetryOptions = {
+			retries: CloudOperationV1.OPERATION_STATUS_CHECK_RETRY_COUNT,
+			minTimeout: CloudOperationV1.OPERATION_STATUS_CHECK_INTERVAL
+		};
+		await promiseRetry(async (retry, attempt) => {
+			try {
+				try {
```
tsvetie: Can we remove one try ... catch block?

lib/s3-helper.ts
```diff
@@ -0,0 +1,13 @@
+export class S3Helper implements IS3Service {
```
tsvetie: IS3Service => IS3Helper, I guess

lib/commands/interactive-cloud-command.ts
TsvetanMilanov: Hypothetically - yes. But since we are going to remove each handler, I prefer not to add additional logic to handle this hypothetical scenario.

lib/services/cloud-project-service.ts
TsvetanMilanov: Because the initial logic was not `{ silent: true }`. Now we can print the result in the first FOR cycle.

lib/s3-helper.ts
TsvetanMilanov: you-are-right

lib/cloud-operation/cloud-operation-v1.ts
TsvetanMilanov: Done.

lib/services/cloud-service.ts
TsvetanMilanov: Yes. I am removing it in the executeCloudOperation method now.

TsvetanMilanov:
@cla-bot check

TsvetanMilanov:
@cla-bot check

TsvetanMilanov:
@cla-bot check

etabakov:
@cla-bot check

PR #161, opened 2019-03-22
Ping V2

TsvetanMilanov:
Implement cloud operation ping v2


server-configs/config.json
```diff
@@ -1,29 +1,23 @@
 {
-	"apiVersion": "v4",
+	"apiVersion": "pr179",
```
Mitko-Kerezov: This file's changes should be reverted

PR #169, opened 2019-05-16
feat: support only webpack workflow

Fatme:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nativescript-cloud",
-  "version": "1.17.6",
+  "version": "1.17.7-2019-06-18-01",
```
KristianDD: 1.18.0

lib/services/cloud-platform-service.ts
```diff
@@ -0,0 +1,120 @@
...
+	constructor(
+		private $injector: IInjector,
+		private $staticConfig: IStaticConfig
+	) {
+		const cliVersion = this.$staticConfig.version;
+		this.shouldUseOldPlatformService = semver.valid(cliVersion) && semver.lt(cliVersion, semver.prerelease(cliVersion) ? "5.4.0-2019-05-16-13277" : "6.0.0");
+	}
+
+	public get $nsCloudBuildCommandHelper(): IBuildCommandHelper {
+		return this.$injector.resolve("nsCloudBuildCommandHelper");
```
KristianDD: Why we don't let the injector resolve this dependency?

lib/services/cloud-platform-service.ts
```diff
@@ -0,0 +1,120 @@
...
+			const $platformService = this.$injector.resolve("platformService");
+			return $platformService.shouldInstall(device, projectData, config, config.outputPath);
+		}
+
+		const $deviceInstallAppService = this.$injector.resolve("deviceInstallAppService");
+		return $deviceInstallAppService.shouldInstall(device, config);
+	}
+
+	public async deployPlatform(platform: string, outputDirectoryPath: string): Promise<void> {
+		const $deployCommandHelper = this.$injector.resolve("deployCommandHelper");
```
KristianDD: Why can't we leave the injector to resolve this in the constructor. We seem to use different method, but the same helper.

lib/services/cloud-platform-service.ts
Fatme: Because there is a cyclic dependency.

PR #170, opened 2019-05-31
fix: tables output is not correct

rosen-vladimirov:
Due to latest CLI changes, some logger APIs have been deprecated. Use correct new APIs.
Also using `logger.out` prints tables incorrectly - use `logger.info` which is the correct way now.
NOTE: For example output of `tns account` was not showing table at all. This is fixed now.

### DIFF
--- a/lib/commands/account/account-list.ts
+++ b/lib/commands/account/account-list.ts
@@ -30,7 +30,7 @@ export class AccountListCommand extends AccountCommandBase implements ICommand {
 			output = table.toString();
 		}
 
-		this.$logger.out(output);
+		this.$logger.info(output);
 	}
 }
 
--- a/lib/services/cloud-service.ts
+++ b/lib/services/cloud-service.ts
@@ -15,6 +15,7 @@ export abstract class CloudService extends EventEmitter implements ICloudService
 		protected $fs: IFileSystem,
 		protected $httpClient: Server.IHttpClient,
 		protected $logger: ILogger,
+		private $constants: IDictionary<any>,
 		private $nsCloudOperationFactory: ICloudOperationFactory,
 		private $nsCloudOutputFilter: ICloudOutputFilter,
 		private $nsCloudProcessService: IProcessService) {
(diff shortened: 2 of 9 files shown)


lib/services/cloud-service.ts
```diff
@@ -161,4 +160,16 @@ export abstract class CloudService extends EventEmitter implements ICloudService
...
+
+	private printMessageOnTheSameLine(msg: string): void {
+		const logger: any = this.$logger;
+		// Print the output on the same line to have cool effects like loading indicators.
+		// The cloud process will take care of the new lines.
+		if (logger.printInfoMessageOnSameLine) {
+			// Used in CLI before 6.0.0
+			logger.printInfoMessageOnSameLine(msg);
+		} else {
+			this.$logger.info(msg, {[this.$constants.LoggerConfigData.skipNewLine]: true });
```
TsvetanMilanov: Is this going to print the message on the same line like `printInfoMessageOnSameLine `?

lib/commands/account/account-list.ts
```diff
@@ -30,7 +30,7 @@ export class AccountListCommand extends AccountCommandBase implements ICommand {
 			output = table.toString();
 		}
 
-		this.$logger.out(output);
+		this.$logger.info(output);
```
TsvetanMilanov: Why is `this.$logger.out` not working anymore?

lib/services/cloud-service.ts
rosen-vladimirov: yes

lib/commands/account/account-list.ts
rosen-vladimirov: `logger.out` is deprecated and will be removed in one of the future releases, that's why I'm replacing it with `logger.info`

TsvetanMilanov: Based on your other comment, `this.$logger.info ` will print the message on the same line. I think this is going to be a problem here. Can you confirm that the table will be properly displayed using `this.$logger.info` (it won't start on the same line and it will end with new line)

rosen-vladimirov: `logger.info` will print on the same line only when `{[this.$constants.LoggerConfigData.skipNewLine]: true }` is passed. Here we do not pass it, so `logger.info` will add new line at the end of the passed input (same as the way `logger.out` was working in the past.

rosen-vladimirov: logger.out should have worked correctly and it will work after merging this PR: https://github.com/NativeScript/nativescript-cli/pull/4670

However, using the new API is the preferred way.

PR #180, opened 2019-08-28
Support aab build flag

tsvetie:
1. Add support for `--aab` flag for `tns cloud build`
2. Snooze log polling so that two subsequent polls do not overlap.
3. After the cloud build completes, check if new logs were generated between the last poll and the operation status change.

### DIFF
--- a/lib/cloud-operation/cloud-operation-v1.ts
+++ b/lib/cloud-operation/cloud-operation-v1.ts
@@ -12,6 +12,7 @@ class CloudOperationV1 extends CloudOperationBase implements ICloudOperation {
 	private serverStatus: IServerStatus;
 	private statusCheckInterval: NodeJS.Timer;
 	private logsCheckInterval: NodeJS.Timer;
+	private snoozeLogPoll: Boolean;
 
 	constructor(public id: string,
 		protected serverResponse: IServerResponse,
@@ -79,7 +80,13 @@ class CloudOperationV1 extends CloudOperationBase implements ICloudOperation {
 
 	private pollForLogs(): void {
 		this.logsCheckInterval = setInterval(async () => {
+			if (this.snoozeLogPoll) {
+				return;
+			}
+
+			this.snoozeLogPoll = true;
 			await this.getCloudOperationLogs();
+			this.snoozeLogPoll = false;
 
 			const status = this.serverStatus.status;
 			if (status === CloudOperationBase.OPERATION_COMPLETE_STATUS || status === CloudOperationBase.OPERATION_FAILED_STATUS) {
--- a/lib/cloud-operation/cloud-operation-v1.ts
+++ b/lib/cloud-operation/cloud-operation-v1.ts
@@ -12,8 +12,8 @@ class CloudOperationV1 extends CloudOperationBase implements ICloudOperation {
 	private serverStatus: IServerStatus;
 	private statusCheckInterval: NodeJS.Timer;
 	private logsCheckInterval: NodeJS.Timer;
-	private snoozeLogPoll: Boolean;
-	private hasLogPollCompleted: Boolean;
+	private snoozeLogPoll: boolean;
+	private hasLogPollCompleted: boolean;
 
 	constructor(public id: string,
 		protected serverResponse: IServerResponse,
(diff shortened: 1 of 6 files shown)


lib/services/cloud-build-service.ts
```diff
@@ -135,6 +135,10 @@ export class CloudBuildService extends CloudService implements ICloudBuildServic
 			additionalCliFlags.push("--no-bundle");
 		}
 
+		if (androidBuildData.aab) {
```
TsvetanMilanov: You need to check if the `androidBuildData` is defined because it is optional.

lib/definitions/cloud-build-service.d.ts
```diff
@@ -281,6 +286,10 @@ interface IIOSBuildData extends IBuildForDevice {
 interface ICloudBuildOutputDirectoryOptions extends IOutputDirectoryOptions {
 }
 
+interface ICloudBuildOutputOptions extends IOutputDirectoryOptions {
+	extension?: string;
```
TsvetanMilanov: There is no need to create new interface for this. Add `extension?: string` in the `IOutputDirectoryOptions`

lib/definitions/cloud-build-service.d.ts
```diff
@@ -246,6 +246,11 @@ interface IAndroidBuildData {
 	 * Password of the specified certificate. Required and used only for release builds.
 	 */
 	certificatePassword: string;
+
+	/**
+	 * Android App Bundle (--aab) option.
+	 */
+	aab?: string;
```
TsvetanMilanov: Why is this property `string`? I think it should be `boolean`

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -79,13 +89,20 @@ class CloudOperationV1 extends CloudOperationBase implements ICloudOperation {
...
 	private pollForLogs(): void {
 		this.logsCheckInterval = setInterval(async () => {
-			await this.getCloudOperationLogs();
+			if (this.snoozeLogPoll) {
+				return;
+			}
 
 			const status = this.serverStatus.status;
-			if (status === CloudOperationBase.OPERATION_COMPLETE_STATUS || status === CloudOperationBase.OPERATION_FAILED_STATUS) {
+			const hasCompleted = status === CloudOperationBase.OPERATION_COMPLETE_STATUS || status === CloudOperationBase.OPERATION_FAILED_STATUS;
```
TsvetanMilanov: You can change this to:

```suggestion
			const hasCompleted = this.serverStatus.status !== CloudOperationBase.OPERATION_IN_PROGRESS_STATUS;
```

lib/cloud-operation/cloud-operation-v1.ts
```diff
@@ -54,7 +56,15 @@ class CloudOperationV1 extends CloudOperationBase implements ICloudOperation {
 	protected async waitForResultCore(): Promise<ICloudOperationResult> {
 		return new Promise<ICloudOperationResult>((resolve, reject) => {
 			this.statusCheckInterval = setInterval(async () => {
-				this.serverStatus = await this.$nsCloudS3Helper.getJsonObjectFromS3File<IServerStatus>(this.serverResponse.statusUrl);
+				const status = this.serverStatus.status;
+				if (status !== CloudOperationV1.OPERATION_COMPLETE_STATUS && status !== CloudOperationV1.OPERATION_FAILED_STATUS) {
```
TsvetanMilanov:
```suggestion
				if (status === CloudOperationV1.OPERATION_IN_PROGRESS_STATUS) {
```

lib/cloud-operation/cloud-operation-v1.ts
tsvetie: This is a personal preference, but I would rather check for the completed status values - e.g. if we add another non-completed status, we will have to add it here. But, if you would prefer this, I will change it.

lib/definitions/cloud-build-service.d.ts
tsvetie: Update: IOutputDirectoryOptions is in NS CLI code, so I do not think this is a good idea. Renamed it to ICloudOperationOutputOptions.

